PR #7, opened 2019-03-18
Ensure `Bump::new_chunk` can satisfy next request

alexcrichton:
This commit fixes a bug that was discovered through `wasm-bindgen` in
2.2.0 (maybe introduced in #5? unsure) where a call to `Bump::new_chunk`
may not have been guaranteed to satisfy the next requested allocation,
even though it was expected to be so.

The issue was that if a large layout was allocated when space was
reserved for `ChunkFooter` the allocation request didn't take that into
account. This means that if you requested a 10kb allocation, for
example, then after allocating a 10kb chunk the chunk wouldn't have 10kb
available because the footer was used as part of that 10kb.

The fix implemented in this issue is to ensure that the size is always
increased by the size of `ChunkFooter` if necessary, guaranteeing that
the chunk returned by `Bump::new_chunk` can satisfy the next allocation
if a layout is provided.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -276,14 +276,14 @@ impl Bump {
 
     /// Allocate a new chunk and return its initialized footer.
     ///
-    /// If given, `layouts` is a tuple of the current chunk layout and the
+    /// If given, `layouts` is a tuple of the current chunk size and the
     /// layout of the allocation request that triggered us to fall back to
     /// allocating a new chunk of memory.
-    fn new_chunk(layouts: Option<(Layout, Layout)>) -> NonNull<ChunkFooter> {
+    fn new_chunk(layouts: Option<(usize, Layout)>) -> NonNull<ChunkFooter> {
         unsafe {
             let layout: Layout =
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -276,32 +276,41 @@ impl Bump {
...
+                    // Have a reasonable "doubling behavior" but ensure that if
+                    // a very large size is requested we round up to that.
+                    let size_to_allocate = cmp::max(old_doubled, requested.size());
+
+                    // Handle size/alignment of our allocated chunk, taking into
+                    // account an overaligned allocation if one is required.
+                    // Note that we also add to the size a `ChunkFooter` because
+                    // we'll be placing one at the end, and we need to at least
+                    // satisfy `requested.size()` bytes.
+                    let size = cmp::max(
```
fitzgen: Bah -- I was totally aware of this and even thought I was handling this corner case already. Goes to show how nice it is to have unsafe code encapsulated in utility crates, because if I had to deal with this all the time, my code would be crashy as hell :-p

fitzgen:
>  (maybe introduced in #5? unsure)

Yeah, I think so.

PR #11, opened 2019-03-25
Add an alloc_with function

### LINKED ISSUE #10: Consider offering an API similar to copyless
The copyless crate allows one to allocate a large struct on the heap without first creating it on the stack and then copying it. One should be able to do this with bumpalo too.

If you try to run this code, it will cause a **stack** overflow exception (even though we want the data on the heap):

```rust
fn main() {
    let bump = bumpalo::Bump::new();
    bump.alloc([0; 1_000_000_000]);
}
```

The copyless crate solves this by using on the fact that llvm seems to be able to optimize `x = FOO; memcpy(ptr, &x, sizeof(x));` into `*ptr = FOO`.

On the other hand, llvm does not seem to be able to optimize `x = FOO; ptr = alloc(sizeof(x)); memcpy(ptr, &x, sizeof(x));` in the same way, and instead relies on the stack to store the value of `FOO`.

TethysSvensson:
This adds a functionality similar to the boxext or copyless crates, i.e.
it adds a function to allocate and write values directly to the heap
instead of going through the stack first.

Resolves #10.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -430,12 +430,78 @@ impl Bump {
     /// ```
     #[inline(always)]
     pub fn alloc<T>(&self, val: T) -> &mut T {
+        self.alloc_with(|| val)
+    }
+
+    /// Pre-allocate space for an object in this `Bump`, initializes it using
+    /// the closure, then returns an exclusive reference to it.
+    ///
+    /// Calling `bump.alloc(x)` is essentially equivalent to calling
+    /// `bump.alloc_with(|| x)`. However if you use `alloc_with`, then the
+    /// closure will not be invoked until after allocating space for storing
+    /// `x` on the heap.
+    ///
+    /// This can be useful in certain edge-cases related to compiler
+    /// optimizations. When evaluating `bump.alloc(x)`, semantically `x` is
+    /// first put on the stack and then moved onto the heap. In some cases,
+    /// the compiler is able to optimize this into constructing `x` directly
+    /// on the heap, however in many cases it does not.
+    ///
+    /// The function `alloc_with` tries to help the compiler be smarter. In
+    /// most cases doing `bump.alloc_with(|| x)` on release mode will be
+    /// enough to help the compiler to realize this optimization is valid
+    /// and construct `x` directly onto the heap.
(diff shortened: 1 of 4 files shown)


tests/alloc_with.rs
```diff
@@ -0,0 +1,65 @@
+// All of these alloc_with tests will fail with "fatal runtime error: stack overflow" unless LLVM
+// manages to optimize the stack writes away.
+
+#[test]
+#[cfg(not(debug_assertions))]
```
fitzgen: Instead of writing this `cfg` on every item in this file, you can add this to the top of the file:

```rust
#![cfg(not(debug_assertions))]
```

src/lib.rs
```diff
@@ -430,12 +430,78 @@ impl Bump {
     /// ```
     #[inline(always)]
     pub fn alloc<T>(&self, val: T) -> &mut T {
+        self.alloc_with(|| val)
+    }
+
+    /// Pre-allocate space for an object in this `Bump`, initializes it using
```
fitzgen: Fantastic documentation comment, thank you!

TethysSvensson:
I've pushed a new version with your requested changes.

> Would you mind also making a copy of the benchmark functions that uses `alloc_with` instead of `alloc`? Should be an easy copy-paste followed by a tiny edit.

Done.

The benchmarks seem pretty noisy, but that could just be my laptop being unpredictable.

As far as the data on my laptop shows, the `allocate-big` tests are consistently at least at 5% faster. It looks like it might be be closer to 10%, but my data is way too noisy for me to be sure.

I also added an `allocate-huge` test, because I thought `alloc_with` might do even better there. Surprisingly it looks like the improvement is about the same as for `allocate-big`. If you want, I can remove it again.

TethysSvensson:
Should I also update `.travis.yml` to run the tests on release as well?

fitzgen:
> Should I also update `.travis.yml` to run the tests on release as well?

I can do it in a follow up.

fitzgen:
Published 2.3.0 with this in it! Thanks again @TethysSvensson!

PR #12, opened 2019-04-01
Add `collect_in` for iterators.

derekdreery:
This adds an extension trait for a new iterator method `collect_in`, that allows using `collect` on a bump allocator.

I've made the PR to stimulate discussion, it didn't take very long so feel free to reject if you feel it doesn't fit.

### DIFF
--- a/src/collections/vec.rs
+++ b/src/collections/vec.rs
@@ -84,6 +84,7 @@
 //! [`vec!`]: ../../macro.vec.html
 
 use super::raw_vec::RawVec;
+use crate::Bump;
 use core::cmp::Ordering;
 use core::fmt;
 use core::hash::{self, Hash};
@@ -96,7 +97,6 @@ use core::ops::{Index, IndexMut, RangeBounds};
 use core::ptr;
 use core::ptr::NonNull;
 use std::slice;
-use crate::Bump;
 
 unsafe fn arith_offset<T>(p: *const T, offset: isize) -> *const T {
     p.offset(offset)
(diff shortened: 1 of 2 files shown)


src/collections/vec.rs
```diff
@@ -2339,3 +2346,44 @@ where
         }
     }
 }
+
+/// Similar to `std::iter::FromIterator`, but for a bump vec.
+pub trait FromIteratorBump<'bump, A> {
```
fitzgen: Let's name this `FromIteratorIn` and its method `from_iter_in`

tests/vec.rs
```diff
@@ -9,3 +9,10 @@ fn push_a_bunch_of_items() {
         v.push(x);
     }
 }
+
+#[test]
+fn collect_in_a_bump_arena() {
+    use bumpalo::collections::vec::{IteratorBump, Vec};
+    let b = Bump::new();
+    (0..10_000).collect_in::<Vec<_>>(&b);
```
fitzgen: Let's also test that it contains the items we expect.

derekdreery:
Just out of interest, how will the new collections in std work? Will they take an allocator? Is there some documentation I can read?

fitzgen:
> Just out of interest, how will the new collections in std work? Will they take an allocator? Is there some documentation I can read?

It isn't 100% clear yet ([will probably require a new RFC to sort it all out](https://github.com/rust-lang/rust/issues/42774#issuecomment-477358608)) but the few bits of `std::collections` that are parameterised over an allocator have additional constructors for passing the allocator in by reference.

derekdreery:
Aa I see, so we can replace this when the time comes. I'll just do the tidying up you've suggested.

fitzgen:
Going to close this due to inactivity -- @derekdreery if you want to rebase and do the tidying up mentioned above, I'll be happy to reopen and merge in the future!

PR #16, opened 2019-04-29
Don't spuriously panic on allocations

### LINKED ISSUE #15: Possible unexpected panic from Bump::alloc?
Hello

I was interested in how the thing is working inside, so I read the code. I discovered this thing: https://github.com/fitzgen/bumpalo/blob/master/src/lib.rs#L539.

Now, if I understand it correctly, this can panic „unexpectedly“, even if the user didn't do anything outright wrong.

* Let's say current chunk lives on really high address. For simplicity, let's say it simply lives in the highest page and the end has address 0xFFFFFFFE (so it's even, or something, let's say we are on 32bit).
* The chunk is running out of allocated space, so the finger is also somewhere very close.
* The user asks for allocation that has a sane size, but doesn't fit.

In this case, the address will overflow and panic will happen.

Would it make sense to not panic, but simply return None in that case, because the problem here is simply that the allocation doesn't fit into the current chunk and a new chunk needs to be created?

vorner:
Don't panic if the allocated chunk happens to be close to the upper end
of address space and an allocation, while reasonable sized, would not
fit within the chunk neither the address space. Handle it the same way
as any other chunk-too-small situations.

Fixes #15.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -511,10 +511,6 @@ impl Bump {
     /// The returned pointer points at uninitialized memory, and should be
     /// initialized with
     /// [`std::ptr::write`](https://doc.rust-lang.org/stable/std/ptr/fn.write.html).
-    ///
-    /// ## Panics
-    ///
-    /// Panics if reserving space for `T` would cause an overflow.
     #[inline(always)]
     pub fn alloc_layout(&self, layout: Layout) -> NonNull<u8> {
         if let Some(p) = self.try_alloc_layout_fast(layout) {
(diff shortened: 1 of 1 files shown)


vorner:
The tests seem to complain on my computer about readme not being updated, but I don't believe that is something I've done.

fitzgen:
The test for overflow is OOMing now:

```
thread 'alloc_overflow' panicked at 'assertion failed: !data.is_null()', src/lib.rs:318:13
```

I'll fix up the test locally and merge this PR. Thanks @vorner!

> The tests seem to complain on my computer about readme not being updated, but I don't believe that is something I've done.

You may not have `cargo-readme` installed, or maybe not its latest version. No worries.

PR #20, opened 2019-05-19
Pass correct Layout to dealloc

### LINKED ISSUE #19: Large alignments cause heap corruption
Contrived example:

```
use bumpalo::Bump;
use std::alloc::Layout;

fn main() {
    let bump = Bump::new();
    let layout = Layout::from_size_align(1024, 32).unwrap();
    bump.alloc_layout(layout);
}
```
Blows up with:
```
error: process didn't exit successfully: `target\debug\crash.exe` (exit code: 0xc0000374, STATUS_HEAP_CORRUPTION)
```
This only seems to happen with alignments greater than 16, and only if the allocation is above a certain size (maybe big enough to overflow the current chunk?).

jsheard:
Fixes #19

`dealloc` was always being called with the default chunk layout rather than the one in the chunk footer.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -228,7 +228,7 @@ impl Drop for Bump {
             let mut footer = Some(self.all_chunk_footers.get());
             while let Some(f) = footer {
                 footer = f.as_ref().next.get();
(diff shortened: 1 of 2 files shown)


fitzgen:
Published 2.4.3

PR #29, opened 2019-10-26
Add a with_capacity function

TethysSvensson:


TethysSvensson:
> Can you add a test that creates a `Bump` with a funky capacity (something that is not footer aligned, like `11`) and then also allocate things out of that bump arena that require more alignment (like `u128`s)? At least enough to cause the bump arena to allocate a new chunk.

Yes, I can definitely add a test like that. :)

I have also simplified the code a bit, to make it more obvious that this will never allocate less than a normal call to `Bump::new()` would.

TethysSvensson:
@fitzgen I've now added the test.

Issue #27, opened 2019-10-26
Iterator of allocated chunks


TethysSvensson:
I am considering writing a function `unsafe fn iter_allocated_chunks(&mut self) -> ChunkIterator<'a>` and then re-implementing `each_allocated_chunk` on top of this function.

Would you be interested in accepting a PR that does this?

TethysSvensson:
Implemented in #30.

PR #30, opened 2019-10-27
Replace each_allocated_chunk function with iter_allocated_chunks function

TethysSvensson:
This implements the suggestion in #27.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -723,7 +723,7 @@ impl Bump {
     }
 }
 
-/// An iterator over the each chunk of allocated memory that
+/// An iterator over each chunk of allocated memory that
 /// an arena has bump allocated into.
 ///
 /// Chunks are returned in order of allocation: oldest chunks first, newest
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -721,6 +721,64 @@ impl Bump {
             bump: PhantomData,
         }
     }
+
+    /// Call `f` on each chunk of allocated memory that this arena has bump
+    /// allocated into.
+    ///
+    /// `f` is invoked in order of allocation: oldest chunks first, newest
+    /// chunks last.
+    ///
+    /// ## Safety
+    ///
+    /// Because this method takes `&mut self`, we know that the bump arena
+    /// reference is unique and therefore there aren't any active references to
+    /// any of the objects we've allocated in it either. This potential aliasing
+    /// of exclusive references is one common footgun for unsafe code that we
+    /// don't need to worry about here.
+    ///
+    /// However, there could be regions of uninitialized memory used as padding
+    /// between allocations. Reading uninitialized memory is big time undefined
+    /// behavior!
+    ///
+    /// The only way to guarantee that there is no padding between allocations
+    /// or within allocated objects is if all of these properties hold:
+    ///
+    /// 1. Every object allocated in this arena has the same alignment.
+    /// 2. Every object's size is a multiple of its alignment.
+    /// 3. None of the objects allocated in this arena contain any internal
+    ///    padding.
+    ///
+    /// If you want to use this `each_allocated_chunk` method, it is *your*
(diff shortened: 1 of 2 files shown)


TethysSvensson:
Note that this currently removes the function `each_allocated_chunk`, since I believe this new functionality to be strictly superior. The iterator API is more usable, while also being safe to use as long as you do not look at any of the byte values. The type signature makes it clear exactly which parts are unsafe.

I can easily re-implement the old function on top of the iterator for backwards-compatiblity, if you prefer. If you want me to do this, I think we should consider marking as deprecated, or perhaps changed to use `[MaybeUninit<u8>]` slices instead of `[u8]` slices (though that you also be a breaking change).

TethysSvensson:
By using google and github search, I have not been able to find any references to `each_allocated_chunk` outside of this crate or rustdocs. There might be users of the function in private or non-indexed code though.

fitzgen:
> By using google and github search, I have not been able to find any references to `each_allocated_chunk` outside of this crate or rustdocs. There might be users of the function in private or non-indexed code though.

I use it here, fwiw: https://github.com/fitzgen/dodrio/blob/22faf1eb60f9bcb16094e1b2f4575edfb203338e/src/change_list/emitter.rs#L43

src/lib.rs
```diff
@@ -663,10 +667,10 @@ impl Bump {
         }
     }
 
-    /// Call `f` on each chunk of allocated memory that this arena has bump
-    /// allocated into.
+    /// Returns an iterator over the each chunk of allocated memory that
```
fitzgen:
```suggestion
    /// Returns an iterator over each chunk of allocated memory that
```

src/lib.rs
```diff
@@ -663,10 +667,10 @@ impl Bump {
...
+    /// from an arena has bump allocated into.
```
fitzgen:
```suggestion
    /// this arena has bump allocated into.
```

src/lib.rs
```diff
@@ -705,20 +710,39 @@ impl Bump {
...
-            let foot = foot.as_ref();
+    pub fn iter_allocated_chunks(&mut self) -> ChunkIter<'_> {
+        ChunkIter {
+            footer: Some(self.all_chunk_footers.get()),
+            bump: PhantomData,
+        }
+    }
+}
 
+/// An iterator over the each chunk of allocated memory that
```
fitzgen:
```suggestion
/// An iterator over each chunk of allocated memory that
```

src/lib.rs
```diff
@@ -705,20 +710,39 @@ impl Bump {
...
+/// from an arena has bump allocated into.
```
fitzgen:
```suggestion
/// an arena has bump allocated into.
```

src/lib.rs
```diff
@@ -705,20 +710,39 @@ impl Bump {
...
+///
+/// Chunks are returned in order of allocation: oldest chunks first, newest
+/// chunks last.
+///
+/// This struct is created by the [`iter_allocated_chunks`] method on
+/// [Bump]. See that function for a safety description regarding.
```
fitzgen:
```suggestion
/// [Bump]. See that function for a safety description regarding reading from the returned items.
```

TethysSvensson:
> > By using google and github search, I have not been able to find any references to `each_allocated_chunk` outside of this crate or rustdocs. There might be users of the function in private or non-indexed code though.
> 
> I use it here, fwiw: https://github.com/fitzgen/dodrio/blob/22faf1eb60f9bcb16094e1b2f4575edfb203338e/src/change_list/emitter.rs#L43

That also shows up on github search now, but didn't before. It still does not show up on google search. This makes me a lot less inclined to trust github search in the future.


> Can you also reimplement `each_allocated_chunk` in terms of `iter_allocated_chunks` with a `#[deprecated(note = "deprecated in favor of iter_allocated_chunks")]` note?

Done.

PR #36, opened 2019-10-30
Reserve space in Vec::extend() before pushing elements

andylokandy:
Reserve space by size_hint() in Vec::extend() before pushing elements.


TethysSvensson:
As far as I know, standard Vec is using all sorts of specialization magic to optimize it's extend and we are not.

fitzgen:
Yeah that's fair.

Issue #41, opened 2019-11-05
Provide a method to allocate a slice of zeroed/default values


ztlpn:
Currently to bump-allocate any slice an existing slice is needed which is not always convenient. I guess currently the workaround is to use the `alloc_layout` method, but a method that doesn't require `unsafe` would be handy. Another handy addition would be a method that allocates and fills a slice with copies/clones of a single value.

ztlpn:
@fitzgen I can do a PR if you think that this addition makes sense.

fitzgen:
Can you sketch the API / method signature you're looking for here?

Zeroing memory is not going to result in a safe API in the general case (hence [`std::mem::zeroed`](https://doc.rust-lang.org/nightly/std/mem/fn.zeroed.html) being `unsafe`) so if your goal is avoiding `unsafe` I think we might need to workshop the API design a bit.

ztlpn:
Something like
```rust
fn alloc_slice_fill_copy<T: Copy>(&self, len: usize, value: &T) -> &mut [T];
fn alloc_slice_fill_clone<T: Clone>(&self, len: usize, value: &T) -> &mut [T];
fn alloc_slice_fill_default<T: Default>(&self, len: usize) -> &mut [T];
```
Basically the equivalent of `vec![len; value]`.

> Zeroing memory is not going to result in a safe API in the general case

Yeah, I don't need zeroed memory in particular, but slice with default values. It was just that I had integer types in mind when I created the issue.

fitzgen:
That looks great -- happy to take a PR implementing those APIs!

fitzgen:
I believe this is fully implemted now. If I'm mistaken, please re-open!

ztlpn:
Thanks! @fitzgen It would be great if you could cut a release with this feature!

fitzgen:
See #51

PR #44, opened 2019-11-07
Implement some more slice allocation functions

TethysSvensson:
This implements the feature requested in #41.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -663,10 +663,6 @@ impl Bump {
     where
         F: FnMut(usize) -> T,
     {
-        if len == 0 {
-            return &mut [];
-        }
-
         let layout = layout_for_array::<T>(len).unwrap_or_else(|| oom());
         let dst = self.alloc_layout(layout).cast::<T>();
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -595,6 +615,114 @@ impl Bump {
...
+    /// ## Example
+    ///
+    /// ```
+    /// let bump = bumpalo::Bump::new();
+    /// let x = bump.alloc_slice_fill_copy(5, 42);
+    /// assert_eq!(x, &[42, 42, 42, 42, 42]);
+    /// ```
+    #[inline(always)]
+    #[allow(clippy::mut_from_ref)]
+    pub fn alloc_slice_fill_copy<T: Copy>(&self, len: usize, value: T) -> &mut [T] {
```
ztlpn: Hey @TethysSvensson thanks for implementing the feature I requested, that was quick! One remark: I messed up the order of arguments in the `vec![]` macro in my request (as I always do, it seems impossible to remember!) - the correct order is `vec![value, len]`. So I think here we should use the same order fo consistency.

TethysSvensson: You are right, that makes more sense.

TethysSvensson:
> Implementation looks great -- can you also add some tests for these new methods? Thanks!

Do you mean beyond the ones in the doctests? Which tests would you like me to implement?

fitzgen:
> Do you mean beyond the ones in the doctests? Which tests would you like me to implement?

* tests for the `len == 0` fast path

* tests for when the layout overflows in rounding up and in the multiply

* would be good to get some quickchecks in `tests/quickchecks.rs` as well:

```rust
quickcheck! {
    // ...

    fn alloc_slices(allocs: Vec<(u8, usize)>) -> () {
        let b = Bump::new();
        let mut allocated = vec![];
        for (val, len) in allocs {
            let s = b.alloc_slice_copy(val, len); // maybe need to place a max on `len`.

            // todo: assert that the slice is all `val`

            let range = (s.as_ptr(), s.as_ptr().add(s.len()));
            for r in &allocated {
                assert!(!overlaps(*r, range));
            }
            allocated.push(range);
        }
    }
}
```

Thanks!

TethysSvensson:
I think we should also provide a function that takes a closure and calls it for every element. That would also be useful internally to avoid code duplication.

fitzgen:
> I think we should also provide a function that takes a closure and calls it for every element. That would also be useful internally to avoid code duplication.

:+1:

TethysSvensson:
I have rebased this on top of master and added a new commit.

TethysSvensson:
I am still missing more tests before this can merge.

TethysSvensson:
@fitzgen I've implemented the tests now. Please take a look.

I also moved the special-casing of ZSTs into the core allocation instead of dealing with it in `alloc_slice`. There is a test that the allocation path works as expected. I am not sure how to test the deallocation/reallocation path. If you have any ideas, I'll be happy to add tests, but otherwise I think the code is simple enough that it works as is.

PR #47, opened 2019-11-09
Fixes #46 by adding a end_of_first_allocation field

### LINKED ISSUE #46: alloc_layout_slow and guarantees for iter_allocated_chunks
I agree that the fix made in #45 is the right one, however it has a problem: It can introduce padding bytes in the output of `iter_allocated_chunks` for large alignments.

I think it is fine to introduce padding for large alignments, but we need to decide where to draw the line between support and not supported -- and we need to improve the documentation to reflect this.

I do not have any strong opinions on how large an alignment we should support. I guess it would be nice to support up to alignment 16, since that is largest alignment of that real world code can sometimes need.

However I do not think that anybody who has 16-byte alignment requirements would also want to use `iter_allocated_chunks`.

Here is some code I used to test this:

```rust
const SUPPORTED_ALIGNMENTS: &[usize] = &[1, 2, 4, 8];

quickcheck::quickcheck! {
    fn test_alignment_chunks(sizes: Vec<usize>) -> () {
        for &alignment in SUPPORTED_ALIGNMENTS {
            let mut b = Bump::new();
            let mut sizes = sizes.iter().map(|&size| (size % 10) * alignment).collect::<Vec<_>>();

            for &size in &sizes {
                let layout = std::alloc::Layout::from_size_align(size, alignment).unwrap();
                let ptr = b.alloc_layout(layout).as_ptr() as *const u8 as usize;
                assert_eq!(ptr % alignment, 0);
            }

            for chunk in b.iter_allocated_chunks() {
                let mut remaining = chunk.len();
                while remaining > 0 {
                    let size = sizes.pop().expect("too many bytes in the chunk output");
                    assert!(remaining >= size, "returned chunk contained padding");
                    remaining -= size;
                }
            }
            assert_eq!(sizes.into_iter().sum::<usize>(), 0);
        }
    }
}
```

TethysSvensson:
This implements the suggested fix in #46.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -188,6 +188,11 @@ struct ChunkFooter {
 
     // Bump allocation finger that is always in the range `self.data..=self`.
     ptr: Cell<NonNull<u8>>,
+
+    // Pointer to the end of the first allocation made in this chunk.
+    // Used in iter_allocated_chunks to avoid giving back padding bytes
+    // that are outside the user's control
+    end_of_first_allocation: Cell<Option<NonNull<u8>>>,
 }
 
 impl Default for Bump {
@@ -352,6 +357,7 @@ impl Bump {
                     layout,
                     prev: Cell::new(prev),
                     ptr,
+                    end_of_first_allocation: Cell::new(None),
                 },
             );
 
@@ -403,6 +409,7 @@ impl Bump {
 
             // Reset the bump finger to the end of the chunk.
             cur_chunk.as_ref().ptr.set(cur_chunk.cast());
+            cur_chunk.as_ref().end_of_first_allocation.set(None);
 
             debug_assert!(
                 self.current_chunk_footer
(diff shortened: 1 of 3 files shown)


TethysSvensson:
I'm worried that this might be too complicated an implementation, or the check in `try_alloc_layout_fast` it might be detrimental to performance.

I will try to see how difficult it is to instead always guarantee that alignments up to and including 16 bytes will work.

src/lib.rs
```diff
@@ -606,15 +613,25 @@ impl Bump {
...
 
-            let ptr = ptr.checked_sub(layout.size())?;
+            let ptr = initial_ptr.checked_sub(layout.size())?;
             let aligned_ptr = ptr & !(layout.align() - 1);
 
             if aligned_ptr >= start {
+                // We might be the very first allocation made in the entire bump,
+                // in which case we need to set end_of_first_allocation
+                // for the first chunk
+                if initial_ptr == footer as *const ChunkFooter as usize {
```
fitzgen: I was hoping we would be able to avoid the check in the fast path and only do it in the slow path, but I realize now that we can't because of the very first allocation. Blech.

fitzgen:
Agreed that the other approach is better.

PR #48, opened 2019-11-09
Fixes for #46 by always aligning the requested memory to 16 bytes

TethysSvensson:
Before this commit, it was not safe to use iter_allocated_chunks if you
had allocated objects with a larger alignment that size_of::<usize>(),
even though the documentation did not mention this. #46 is about that bug.

This commit chooses to partially remedy this problem, and partially
update the documentation.

Specifically this commit fixes the code, so that alignments up to and
included 16 bytes are supported, but larger alignments will still cause
problems. It also updates the documentation, to specify that larger
alignments are unsupported when used together with
iter_allocated_chunks.

We do not believe this to be an issue, since we do not believe that
anybody using bumpalo will have alignment requirements larger than 16
bytes and also want to use iter_allocated_chunks.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -959,7 +959,7 @@ mod tests {
         use crate::alloc::Alloc;
 
         unsafe {
-            const CAPACITY: usize = 1024;
+            const CAPACITY: usize = 1024 - OVERHEAD;
             let mut b = Bump::with_capacity(CAPACITY);
 
             // `realloc` doesn't shrink allocations that aren't "worth it".
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -286,9 +286,7 @@ impl Bump {
     /// ```
     pub fn with_capacity(capacity: usize) -> Bump {
         let chunk_footer = Self::new_chunk(
-            // We divide by 2, since this is the "old size", which will
-            // be multiplied by two inside the new_chunk function
-            DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER / 2,
+            None,
             Some(unsafe { layout_from_size_align(capacity, 1) }),
             None,
         );
@@ -303,19 +301,32 @@ impl Bump {
     /// layout of the allocation request that triggered us to fall back to
     /// allocating a new chunk of memory.
     fn new_chunk(
-        old_size_without_footer: usize,
+        old_size_with_footer: Option<usize>,
         requested_layout: Option<Layout>,
         prev: Option<NonNull<ChunkFooter>>,
     ) -> NonNull<ChunkFooter> {
         unsafe {
             // We want to at least double the available size on every chunk allocation
-            let mut new_size_without_footer = old_size_without_footer.checked_mul(2).unwrap();
-            debug_assert_eq!(
-                new_size_without_footer,
-                round_up_to(new_size_without_footer, CHUNK_ALIGN).unwrap(),
-                "The old size was already a multiple of our chunk alignment, so no \
-                 need to round it up again."
-            );
+            let mut new_size_without_footer =
+                if let Some(old_size_with_footer) = old_size_with_footer {
+                    // This code ensure that we roughly double every time, while
+                    // fulfilling alignment requirements. We also prefer to request just
+                    // short of a full number of pages, instead of just above a full number
+                    // of pages. On 64-bit code means that quickly begin requesting chunks
+                    // with size 0x???ff8, while on 32-bit we begin requesting chunks with
+                    // size 0x???ff4.
+                    round_up_to(
+                        old_size_with_footer
+                            .checked_mul(2)
+                            .unwrap_or_else(|| oom())
+                            .checked_sub(FOOTER_SIZE)
+                            .unwrap(),
+                        CHUNK_ALIGN,
+                    )
+                    .unwrap_or_else(|| oom())
+                } else {
+                    DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER
+                };
 
             // We want to have CHUNK_ALIGN or better alignment
             let mut align = CHUNK_ALIGN;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -340,6 +333,12 @@ impl Bump {
                 new_size_without_footer = new_size_without_footer.max(requested_size);
             }
 
+            // If we are in the range of using entire pages, let's actually use entire pages
+            if new_size_without_footer >= PAGE_STRATEGY_CUTOFF {
+                new_size_without_footer =
```
fitzgen: This will work well if the allocator is storing its metadata about allocations in a side table, but if it is storing them in a header or footer, we are going to get something akin to [clown shoes in sizes 2^10+1 and up](https://blog.mozilla.org/nnethercote/2011/08/05/clownshoes-available-in-sizes-2101-and-up/). That is, the allocator needs a little space to keep track of these allocations, so if the allocation is perfectly multiples-of-page-sized, we can end up allocating an extra page that is almost all unused.

If we are going to go to the trouble of rounding to multiples of a page, lets also remove `MALLOC_OVERHEAD` from the total size (if we can without going under the requested allocation size), which should alleviate this problem.

src/lib.rs
```diff
@@ -225,14 +224,29 @@ pub(crate) fn round_up_to(n: usize, divisor: usize) -> Option<usize> {
...
 }
 
+// We need a const fn for max
+const fn max(a: usize, b: usize) -> usize {
+    [a, b][(a < b) as usize]
+}
+
+// We only support alignments of up to 16 bytes for iter_allocated_chunks.
+const SUPPORTED_ITER_ALIGNMENT: usize = 16;
+const CHUNK_ALIGN: usize = max(SUPPORTED_ITER_ALIGNMENT, mem::align_of::<ChunkFooter>());
```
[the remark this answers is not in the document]
TethysSvensson: I don't think I understand this. Shouldn't it be const?

src/lib.rs
TethysSvensson: Please take another look. I think I also managed to make the code simpler.

According to the page you linked, we should potentially start rounded to a megabyte bounded after we hit the first megabyte as well.

TethysSvensson: During normal use, we will hit 1m alignment by just doubling, but if a user does either `with_capacity` or allocates something very large, we might only be 4k aligned instead of 1m aligned.

src/lib.rs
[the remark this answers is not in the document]
fitzgen: Yeah, the whole thing could be `const` too

src/lib.rs
fitzgen: I don't think we should over-fit to jemalloc's size classes (especially not the exact size classes reported in that post, since that is (a) a few years old, and (b) specific to mozjemalloc, a fork of jemalloc from even a few years earlier than that)

I care quite a bit about working well with dlmalloc (the default on most "interesting", non-mainstream targets such as wasm32-unknown-unknown) as well as wee_alloc.

I think the page rounding should be a happy place for all of them.

TethysSvensson: Sounds good.

fitzgen:
the new commits look good to me -- nice clean up!

`realloc` test seems to be failing:

```
thread 'tests::test_realloc' panicked at 'assertion failed: q.as_ptr() as usize != p.as_ptr() as usize - CAPACITY', src/lib.rs:992:13
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:200
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:477
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:384
   8: rust_begin_unwind
             at src/libstd/panicking.rs:311
   9: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  10: core::panicking::panic
             at src/libcore/panicking.rs:49
  11: bumpalo::tests::test_realloc
             at src/lib.rs:992
  12: bumpalo::tests::test_realloc::{{closure}}
             at src/lib.rs:959
  13: core::ops::function::FnOnce::call_once
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54/src/libcore/ops/function.rs:235
  14: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54/src/liballoc/boxed.rs:787
  15: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:80
  16: std::panicking::try
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54/src/libstd/panicking.rs:275
  17: std::panic::catch_unwind
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54/src/libstd/panic.rs:394
  18: test::run_test::run_test_inner::{{closure}}
             at src/libtest/lib.rs:1408
```

src/lib.rs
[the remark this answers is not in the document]
TethysSvensson: So how about I do this then?

```rust
// We only support alignments of up to 16 bytes for iter_allocated_chunks.
const SUPPORTED_ITER_ALIGNMENT: usize = 16;
const CHUNK_ALIGN: usize = SUPPORTED_ITER_ALIGNMENT;

// Assert that ChunkFooter is at most the supported alignment. This will give a compile time error if it is not the case
const _FOOTER_ALIGN_ASSERTION: bool = mem::align_of::<ChunkFooter>() <= CHUNK_ALIGN;
const _: [(); _FOOTER_ALIGN_ASSERTION as usize] = [()];
```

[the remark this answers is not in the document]
TethysSvensson: I have pushed a commit containing basically this.

TethysSvensson:
I have fixed the broken test.

PR #54, opened 2019-12-10
HashMap using RawTable from hashbrown

hansihe:
This is a copy/paste of the `HashMap` implementation from `hashbrown`.

This depends on https://github.com/rust-lang/hashbrown/pull/133, which adds allocator support to `hashbrown`'s `RawTable`. Do not merge before this lands.

This is split into two parts:
1. Add a `unstable_core_alloc` feature that uses the `Alloc` trait in std instead of the copied internal version.
2. Add a `collections_hash` feature that depends on both the `unstable_core_alloc` feature and the `hashbrown` crate. I thought it was best to add a new feature for this since it adds a whole new dependency.

Bikeshedding of the feature names is encouraged. If this is a welcome addition, I would be open to adding `HashSet` as well, which should require significantly less code.


TethysSvensson:
I am also doubtful of whether this would actually be useful in practice, since bumpalo is not well-suited for applications in need of frequent re-allocations.

fitzgen:
Closing this as explained in https://github.com/fitzgen/bumpalo/pull/54#pullrequestreview-332768122

Amanieu:
I'm happy to accept a PR adding `allocator_api` support to hashbrown. You can extend your existing PR (https://github.com/rust-lang/hashbrown/pull/133) to add allocator support to the full `HashMap` API instead of just the internal `RawTable` API.

PR #58, opened 2019-12-24
Add `allocated_bytes` method

### LINKED ISSUE #57: Add method to return currently allocated bytes size?
I'm working with an application that prints diagnostic messages after running an operation to show how much memory was used during that operation, then proceeds on with the next operation. However, because the next operation relies on memory allocated in the previous operation, there are still active immutable borrows from the `Bump` allocator at the point where diagnostic information should be logged. Therefore, I cannot simply use `arena.iter_allocated_chunks().map(|chunk| chunk.len()).sum()` to return the allocated size, because that requires a mutable borrow.

Would you be willing to add a method for querying such information from an immutable `Bump` instance? Something like `Bump::allocated_bytes(&self) -> usize`

novacrazy:
Using your advice in #57 made this pretty simple. Thanks!

Fixes #57

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -980,6 +980,35 @@ impl Bump {
         }
     }
 
+    /// Calculates the number of bytes currently allocated across all chunks
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// let bump = bumpalo::Bump::new();
+    /// let _x = bump.alloc_slice_fill_default::<u32>(5);
+    /// let bytes = bump.allocated_bytes();
+    /// assert!(bytes >= core::mem::size_of::<u32>() * 5);
+    /// ```
+    pub fn allocated_bytes(&self) -> usize {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -980,6 +980,35 @@ impl Bump {
         }
     }
 
+    /// Calculates the number of bytes currently allocated across all chunks
+    ///
```
fitzgen: Can you add a snippet here about how this may include padding inserted for alignment?

> If you allocate types of different alignments or types with larger-than-typical alignment in the same arena, some padding bytes might get allocated in the bump arena. Note that those padding bytes will add to this method's resulting sum, so you cannot rely on it *only* counting the sum of the sizes of the things you've allocated in the arena.

novacrazy: Ah, good point. Added.

PR #67, opened 2020-02-14
[WIP] Added try_alloc_layout, try_reserve, & try_reserve_exact

### LINKED ISSUE #66: Add try_reserve/_exact methods for Vec?
I tried to add these try_* methods (since I need fallible allocation) to `Vec` by wrapping the `RawVec` methods of the same names. However, they still seem to be infallible due to the alloc impl always returning `Ok`? https://github.com/fitzgen/bumpalo/blob/master/src/lib.rs#L1076

Would it be possible to make them fallible?

TheDan64:
1) Made it a point to not make any breaking changes to the public API.
2) I implemented my own global allocator which always returns a null ptr to simulate OOM and it seemed like `Bump::new()` /  `Bump::with_capacity(0)` would always panic with `oom()`. Shouldn't it not try and reach for the allocator in that case? Any idea how to go about fixing that special case?
3) I named the new methods `try_reserve` and `try_reserve_exact` to match what raw/vec calls them, but I could rename them to `try_alloc_in_new_chunk_if_necessary` like you suggested if you wish.

Fixes #66

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -319,7 +319,7 @@ impl Bump {
         Bump::try_with_capacity(0)
     }
 
-    /// Construct a new arena with the specified capacity to bump allocate into.
+    /// Construct a new arena with the specified byte capacity to bump allocate into.
     ///
     /// ## Example
     ///
@@ -331,7 +331,7 @@ impl Bump {
         Bump::try_with_capacity(capacity).unwrap_or_else(|_| oom())
     }
 
-    /// Attempt to construct a new arena with the specified capacity to bump allocate into.
+    /// Attempt to construct a new arena with the specified byte capacity to bump allocate into.
     ///
     /// ## Example
     ///
@@ -874,8 +874,16 @@ impl Bump {
         }
     }
 
-    // Slow path allocation for when we need to allocate a new chunk from the
-    // parent bump set because there isn't enough room in our current chunk.
+    /// Gets the remaining capacity in the current chunk.
+    pub fn chunk_capacity(&self) -> usize {
+        let current_footer = self.current_chunk_footer.get();
+        let current_footer = unsafe { current_footer.as_ref() };
+
+        current_footer as *const _ as usize - current_footer.data.as_ptr() as usize
+    }
+
+    /// Slow path allocation for when we need to allocate a new chunk from the
+    /// parent bump set because there isn't enough room in our current chunk.
     #[inline(never)]
     fn alloc_layout_slow(&self, layout: Layout) -> Option<NonNull<u8>> {
         unsafe {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -320,7 +320,8 @@ impl Bump {
             None,
             Some(unsafe { layout_from_size_align(capacity, 1) }),
             None,
-        );
+        ).unwrap_or_else(|| oom());
+
         Bump {
             current_chunk_footer: Cell::new(chunk_footer),
         }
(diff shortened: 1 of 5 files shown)


fitzgen:
> I implemented my own global allocator which always returns a null ptr to simulate OOM and it seemed like `Bump::new()` / `Bump::with_capacity(0)` would always panic with `oom()`. Shouldn't it not try and reach for the allocator in that case? Any idea how to go about fixing that special case?

For these cases, we should add `try_with_capacity` and `try_new` variants that return `Result<Self>`. `Bump` isn't designed to exist without a chunk to allocate into, which would require new branches in the allocation fast path.

src/lib.rs
```diff
@@ -335,16 +336,17 @@ impl Bump {
...
             // big as the previous allocation
             let mut new_size_without_footer =
                 if let Some(old_size_with_footer) = old_size_with_footer {
                     let old_size_without_footer = old_size_with_footer - FOOTER_SIZE;
-                    old_size_without_footer
-                        .checked_mul(2)
-                        .unwrap_or_else(|| oom())
+                    match old_size_without_footer.checked_mul(2) {
+                        Some(size) => size,
+                        None => return None,
```
fitzgen: I think this whole match can become `old_size_without_footer.checked_mul(2)?` and use `?` propagation.

fitzgen:
Please leave a comment here when this is ready for review again -- thanks!

TheDan64:
I added the test you roughed out, but it seems to make Rust go OOM in the middle of the loop, not sure why Rust internals are trying to allocate there.

TheDan64:
Did you have a chance to poke around at the test?

fitzgen:
> Did you have a chance to poke around at the test?

Unfortunately I don't have time to look at this PR this week, and probably won't until wednesday or thursday next week.

In the meantime, I was thinking it may make sense to have a specialized `assert!` macro for inside this test file that automatically turns the global allocator on again if the assertion condition triggers, but before the panic actually happens.

TheDan64:
I think a problem that I'm running into is that cargo will run tests in parallel, so multiple `try_alloc.rs` tests will interfere with each other. And disabling parallel tests for the entire test suite seems less than ideal. So maybe something like a mutex or any other lock really is ideal so that each test can be guaranteed exclusive access to the allocator.

TheDan64:
And I'm also not convinced that the separate file separates the allocator from the other existing tests. I did see one tests randomly go OOM:
`thread 'alloc_slice_overflow' panicked at 'out of memory', src/lib.rs:1104:5`

TheDan64:
Hey, @fitzgen, any chance you'll have an opportunity to look at this? Would like to see this to completion :)

fitzgen:
I don't have time to investigate further myself right now -- if you hammer out the kinks, I'm happy to review tho!

TheDan64:
@fitzgen I think this is ready for final review now. Unfortunately I had to combine my two tests into one, as tests are run in parallel by default and so the global allocator would toggle during one test, potentially affecting the other. Using a Mutex in the allocator doesn't help because it needs to be locked for the entirety of a single test, and Rust accesses it "async"ly. Hypothetically this could be accomplished with a thread-local allocator, but this isn't supported by Rust (not sure if that's a thing in general).

fitzgen:
Thanks!

fitzgen:
FWIW, I split the tests up again by writing a custom `main` function, rather than using the default test harness. This is usually overkill, but I found that the tests were super fragile without it (as we also saw here with the difficulty of getting it green in the first place).

fitzgen:
Also, I published 3.30 with these changes.

PR #68, opened 2020-02-29
Update `Alloc` trait to nightly, add nightly feature

hansihe:
* Rename `Alloc` to `AllocRef`, along with latest nightly
* Remove "common usage" functions from the `AllocRef` trait (https://github.com/rust-lang/rust/pull/69026)
* Add `nightly` feature, which uses the alloc trait in core instead of the copied one in the crate


fitzgen:
It looks like `AllocRef` isn't found in nightly CI: https://dev.azure.com/fitzgen/bumpalo/_build/results?buildId=161&view=logs&j=4348da94-6ff4-557c-1304-492c812b9157&t=db45a008-f7f6-5c94-4ab6-81ac5572c26f&l=224

Looks like we're getting rustc 1.41 from rustup: https://dev.azure.com/fitzgen/bumpalo/_build/results?buildId=161&view=logs&j=4348da94-6ff4-557c-1304-492c812b9157&t=a4ec2f30-9a1e-5354-6437-3d76911eae85&l=12

Is `AllocRef` in 1.41?

hansihe:
I meant to get back to this earlier when I saw that the build failed.

It looks like `--default-toolchain $RUSTUP_TOOLCHAIN` might not be respected in https://github.com/fitzgen/bumpalo/blob/master/ci/install-rust.yml#L4? Even though `nightly` is passed to `rust_version`, stable is installed.

I'll poke around a bit

hansihe:
Okey, so it seems like the alloc trait is under major flux right this moment. 

I guess it would make sense to wait until things have settled a bit to update and merge?

I presume what you want to do is to track the (feature flagged) alloc trait present in the stable releases?

hansihe:
Yeah, let's hold off on this until most changes to the `AllocRef` (formerly alloc) trait are out of the way.

https://github.com/rust-lang/hashbrown/pull/133#issuecomment-595413077

fitzgen:
Sounds like a plan!

NOBLES5E:
> Sounds like a plan!

`AllocRef` is on nightly now. Let's start working on merging this? :blush:

fitzgen:
> `AllocRef` is on nightly now. Let's start working on merging this? blush

Sounds good. Unclear to me whether this PR should be revived or whether starting from scratch would be easier. I don't have time to drive it forward, but I can review the eventual implementation.

NOBLES5E:
> > `AllocRef` is on nightly now. Let's start working on merging this? blush
> 
> Sounds good. Unclear to me whether this PR should be revived or whether starting from scratch would be easier. I don't have time to drive it forward, but I can review the eventual implementation.

It seems that this PR only misses the latest two commits (which only contains some modification in the readme file).

@hansihe What do you think? Let's revive this PR?

hansihe:
Both `Box` and `Vec` in nightly have an allocator parameter now. `hashbrown` also has an allocator parameter on both `HashMap` and `HashSet`, and nightly rust should land those allocator parameters soon.

For me at least, things in the Rust ecosystem have progressed to the point where I would be fine with `bumpalo` just providing the `Bump` arena, along with an `AllocRef` implementation.

My fork, located [here](https://github.com/hansihe/bumpalo), updates the traits to latest nightly, but also removes all collections from the codebase.

It's understandable if @fitzgen would prefer to keep the collections in the crate for now. If that's the case, I will be maintaining my fork as the effort is pretty minimal. If someone else would like to contribute an updated collections implementation on top of my fork in order to get things merged back here, I would be really happy with that too.

fitzgen:
Yes, I'd like to keep the `collections` module until the custom allocators ship on stable.

In the meantime, I'd still like to have an `AllocRef`/`Allocator`/whatever-it-ends-up-being-called implementation behind an unstable cargo flag. If we can lift that from your fork, that would be great.

> I would be fine with `bumpalo` just providing the `Bump` arena, along with an `AllocRef` implementation.

Yep, this is the eventual goal but we need custom allocators on stable before we remove the collections module.

(I don't mind you maintaining a fork, or course, but out of curiosity: what would your fork have over the main crate other than removing the collections module? Since the collections module is off by default and requires a cargo flag to enable, it doeTsn't seem like forking is really getting you that much, unless there are additional changes.)

hansihe:
I just don't have the time right now to update the different collections to the newest version of the trait. If this repo lands and update to the trait, then that's brilliant and I'll of course switch to using that directly instead.

fitzgen:
Fixed in #92.

Issue #72, opened 2020-04-13
Implement `Sync` for `Vec`.


zakarumych:
It seems that for `T: Sync` `Vec<'bump, T>`  should implement `Sync` as well.

fitzgen:
I think you're correct. The bump arena is not sync, but `Vec<'bump, T>` won't touch the bump arena unless you have mutable access, which you can't across threads.

Are you interested in submitting a PR?

zakarumych:
Returning to the issue.
There is a way to send `&Bump` to another thread is `Vec` would be `Sync`. So this would be unsound.

PR #76, opened 2020-05-27
Box for bumpalo

zakarumych:
Most of the code is a copy of Rust's `alloc::boxed::Box`.
New type is behind feature "boxed".

Also adds `Vec::into_boxed_slice` and `From<Vec<T>> for Box<[T]>`
when both "boxed" and "collections" features are enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,12 @@ pointer back to the start of the arena's memory chunk. This makes mass
 deallocation *extremely* fast, but allocated objects' `Drop` implementations are
 not invoked.
 
+[`bumpalo::boxed::Box`] can be used to wrap values allocated in the `Bump` arena and
+call drop when the wrapper goes out of scope. Similar to how [`std::boxed::Box`] works.
+
+[`bumpalo::boxed::Box`]: ./boxed/struct.Box.html
+[`std::boxed::Box`]: https://doc.rust-lang.org/std/boxed/struct.Box.html
+
 ### What happens when the memory chunk is full?
 
 This implementation will allocate a new memory chunk from the global allocator
@@ -90,6 +96,28 @@ for i in 0..100 {
 Eventually [all `std` collection types will be parameterized by an
 allocator](https://github.com/rust-lang/rust/issues/42774) and we can remove
 this `collections` module and use the `std` versions.
+
+### Boxed
+
+When the `"boxed"` cargo feature is enabled, a fork of `std::boxed::Box`
+library is available in the `boxed` module. This `Box` type is modified to allocate
+its space inside `bumpalo::Bump` arenas.
+
+```rust
+use bumpalo::{Bump, boxed::Box};
+
+// Create a new bump arena.
+let bump = Bump::new();
+
+// Create a boxed `String` whose storage is backed by the bump arena. The
+// box cannot outlive its backing arena, and this property is enforced with
+// Rust's lifetime rules.
+let mut s = Box::new_in(String::from("The quick brown fox jumps over the lazy dog"), &bump);
+
+// Push a char at the end.
+s.push('!');
+
+```
 
 ### `#![no_std]` Support
 
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -423,18 +423,19 @@ impl<'a, T: ?Sized> Box<'a, T> {
     /// assert_eq!(*reference, 42);
     /// ```
     ///
-    #[cfg_attr(feature = "collections", doc = r##"
-    ```
-    use bumpalo::{Bump, boxed::Box, vec};
-
-    let b = Bump::new();
-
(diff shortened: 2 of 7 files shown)


fitzgen:
Hi! Thanks for this PR, I've wanted this for a while, but have never gotten around to it. I'll dig into the code in a little bit.

@jorendorff, I know that jsparagus has a similar thing, would merging this PR allow you to remove your bespoke `Box` and use this instead? (if not, why not?)

src/boxed.rs
```diff
@@ -0,0 +1,739 @@
...
+    /// ```
+    ///
+    /// [zeroed]: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.zeroed
+    pub fn new_zeroed_in(a: &'a Bump) -> Box<'a, mem::MaybeUninit<T>> {
+        Box(a.alloc_with(|| mem::MaybeUninit::zeroed()))
+    }
+
+    /// Constructs a new `Pin<Box<T>>`. If `T` does not implement `Unpin`, then
+    /// `x` will be pinned in memory and unable to be moved.
+    #[inline(always)]
```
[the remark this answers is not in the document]
zakarumych: Yeap

src/boxed.rs
```diff
@@ -0,0 +1,739 @@
...
+    ///     values[0].as_mut_ptr().write(1);
+    ///     values[1].as_mut_ptr().write(2);
+    ///     values[2].as_mut_ptr().write(3);
+    ///
+    ///     values.assume_init()
+    /// };
+    ///
+    /// assert_eq!(*values, [1, 2, 3])
+    /// ```
+    pub fn new_uninit_slice_in(len: usize, a: &'a Bump) -> Box<'a, [mem::MaybeUninit<T>]> {
```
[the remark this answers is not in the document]
zakarumych: Maybe put them under `#[cfg(feature = "unstable")]`?

src/collections/raw_vec.rs
```diff
@@ -56,6 +56,10 @@ pub struct RawVec<'a, T> {
     a: &'a Bump,
 }
 
+/// `RawVec<T>` in `Sync` if `T` is.
+/// This is valid as `RawVec` doesn't access inner `Bump` in `&self` methods.
+unsafe impl<'a, T> Sync for RawVec<'a, T> where T: Sync {}
```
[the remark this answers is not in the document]
zakarumych: Sorry, I forgot to remove this code, when I was trying to make `Vec<T>: Sync where T: Sync` and found it is in fact unsound, because clonned `Vec<T>` may access `Bump`.
I explained it here #72 as well.

zakarumych:
There's handy feature for doc generation available on nightly, which allows marking parts of API as feature-gated. Would you consider running doc-test on nightly?
Otherwise I have to remove it.

src/collections/raw_vec.rs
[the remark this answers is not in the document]
fitzgen: Ah, thanks I didn't see the code example in that issue originally, and now that I've seen it, it's very clear to me.

README.md
```diff
@@ -91,6 +97,28 @@ Eventually [all `std` collection types will be parameterized by an
...
 this `collections` module and use the `std` versions.
 
+### Boxed
+
+When the `"boxed"` cargo feature is enabled, a fork of `std::boxed::Box`
+library is available in the `boxed` module. This `Box` type is modified to allocate
+its space inside `bumpalo::Bump` arenas.
+
+```rust
+use bumpalo::{Bump, boxed::Box};
```
fitzgen, line 107: Is this the doc example you need the feature for?

We can do something like https://github.com/rust-fuzz/arbitrary/blob/master/src/unstructured.rs#L45 and that way keep the doc tests passing on all channels.

Thanks for adding this example!

zakarumych: No, no. This stuff
https://github.com/fitzgen/bumpalo/pull/76/files#diff-6616a4c448a262bac3046bd5e1ac68cbR91

fitzgen: Is there a reason we can't do the same thing that I linked to in those locations, instead of those nightly-only attributes?

zakarumych: Those different things entirely.
This unstable attribute adds nice badge to the types and functions so that doc reader immediately see what features should be enabled instead of searching this info in the description or, even worse, open code to see that type is behind a feature after compilation fails.

See this module for example.
https://docs.rs/tokio/0.2.21/tokio/fs/index.html

fitzgen: Ah, I see. I think it is fine to leave that information to normal comments, until the attributes are available on all channels.

zakarumych: Ok. Then I remove those attributes.

src/boxed.rs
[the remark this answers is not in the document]
zakarumych: Ok

PR #78, opened 2020-06-04
Use Layout::array and add an MSRV policy

TethysSvensson:


fitzgen:
Looks like CI is still pulling down 1.43; I'll take a look at it later.

fitzgen:
Merged in https://github.com/fitzgen/bumpalo/pull/79

fitzgen:
Thanks!

PR #86, opened 2020-09-26
Add collect_in for iterators

cuviper:
Inspired by `Iterator::collect` and `FromIterator`, this adds extension
trait `BumpIterator` with `collect_in`, using a new `FromIteratorIn`
trait for collections to implement.

This unifies the existing inherent `from_iter_in` methods in each
container, and also allows generic tricks like collecting `Result` or
`Option` with short-circuiting.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -1,0 +1,96 @@
+//! Additional `Bump`-related functionality for iterators.
+
+use crate::Bump;
+
+/// `Iterator` extensions for `Bump`-allocated types.
+pub trait BumpIterator: Iterator {
+    /// Transforms an iterator into a `Bump`-allocated collection.
+    fn collect_in<'bump, C>(self, bump: &'bump Bump) -> C
+    where
+        Self: Sized,
+        C: FromIteratorIn<'bump, Self::Item>,
+    {
+        C::from_iter_in(self, bump)
+    }
+}
+
+impl<I: Iterator> BumpIterator for I {}
+
+/// Conversion from an iterator to a `Bump`-allocated type.
+/// This trait is an adapted version of `FromIterator`.
+///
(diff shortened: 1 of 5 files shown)


cuviper:
This is a draft because it lacks testing, but I'd like feedback on the idea before I spend more time on it.

src/iter.rs
```diff
@@ -0,0 +1,96 @@
+//! Additional `Bump`-related functionality for iterators.
+
+use crate::Bump;
+
+/// `Iterator` extensions for `Bump`-allocated types.
+pub trait BumpIterator: Iterator {
```
cuviper, line 6: Another idea that could go in this trait is `unzip_in(&Bump)`. `Iterator::unzip` requires the collections to implement `Default + Extend` -- maybe there could be a `DefaultIn` trait building from a bump allocator? This would also enable adaptations of `Iterator::partition` and `Itertools::partition_map`.

fitzgen (commented):
This is good stuff, but I'm hesitant to add this sort of thing to `bumpalo` because it keeps spreading to more and more places and touching more and more things. For example the question of whether we would want `DefaultIn`. I *do* think this stuff should exist, but it should exist in `std` and I don't want to personally do all the design work that should ideally be part of the libs team's purview and/or the allocators WG. I think the best way to push this forward would be through those avenues.

cuviper:
That's fine, this was just a passing idea on my part. I don't have the energy to drive ecosystem changes around this.

It's not clear to me whether your hesitance is leaning yes/no, but I'll just close for now. Maybe someone else will come along and argue that they really need it, and then it can always be revived...

fitzgen:
Yes, I think closing is the right choice. I do appreciate you exploring this though!

>  I don't have the energy to drive ecosystem changes around this.

Yeah, I don't really have that energy either. I hope someone does sometime!

PR #91, opened 2021-01-09
Add `try_alloc`, `try_alloc_with`, `alloc_try_with` and `try_alloc_try_with` methods to `Bump`

Tamschi:
(I updated this opening comment to reflect the final state of the pull request. See its edit history for earlier versions.)

- - -

# Changes

This pull requests adds new methods to `Bump` according to the following pattern:

|                     | fallible allocation  | infallible allocation |
|---------------------|----------------------|-----------------------|
| **by value**        | `try_alloc`          | `alloc`¹              |
| **infallible init** | `try_alloc_with`     | `alloc_with`¹         |
| **fallible init**   | `try_alloc_try_with` | `alloc_try_with`      |

¹ already exists

The `try_alloc…` methods supplement the existing `try_alloc_layout` with generic equivalents that can be consumed more flexibly in safe Rust. The `…_try_with` methods take a fallible initialiser and return any `Err` payload it creates by value, also attempt to undo the `Result`'s bump arena allocation in that process.

The pull request also contains relevant tests and benches for each code path (mostly copies of `alloc_with.rs`), but I wasn't sure how to check for the deallocation from outside.

# Motivation

`alloc_try_with` lets consumers use the plain Rust error escalation pattern with `?` when constructing instances in place:

```rust
let b = Bump::new();
b.alloc_try_with(|| {
    // [Fallible code returning an Err somehow]
    Ok([4u8; 10_000_000])
})?;
```

It would be fairly cumbersome and require `unsafe` for consumers to implement this using the plain `alloc_with` method due to the latter's `&mut Result<_, _>` return type with the same kind of closure. It also wouldn't be possible for a consumer to undo the allocation as well, since `Alloc::dealloc` won't free extra padding.

`alloc_try_with` is still a wrapper around `alloc_with`, but exposes a safe API and is thorough with the deallocation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -234,6 +234,7 @@ pub mod collections;
 mod alloc;
 
 use core::cell::Cell;
+use core::fmt::Display;
 use core::iter;
 use core::marker::PhantomData;
 use core::mem;
@@ -243,6 +244,32 @@ use core::str;
 use core_alloc::alloc::{alloc, dealloc, Layout};
 #[cfg(feature = "allocator_api")]
 use core_alloc::alloc::{AllocError, Allocator};
+
+/// An error returned from [`Bump::try_alloc_try_with`].
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum AllocOrInitError<E> {
+    /// Indicates that the initial allocation failed.
+    Alloc(alloc::AllocErr),
+    /// Indicates that the initializer failed with the contained error after
+    /// allocation.
+    ///
+    /// It is possible but not guaranteed that the allocated memory has been
+    /// released back to the allocator at this point.
+    Init(E),
+}
+impl<E> From<alloc::AllocErr> for AllocOrInitError<E> {
+    fn from(e: alloc::AllocErr) -> Self {
+        Self::Alloc(e)
+    }
+}
+impl<E: Display> Display for AllocOrInitError<E> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            AllocOrInitError::Alloc(err) => err.fmt(f),
+            AllocOrInitError::Init(err) => write!(f, "initialization failed: {}", err),
+        }
+    }
+}
 
 /// An arena to bump allocate into.
 ///
@@ -299,6 +326,75 @@ use core_alloc::alloc::{AllocError, Allocator};
 /// let mut s = bump.alloc("bumpalo");
 /// *s = "the bump allocator; and also is a buffalo";
 /// ```
+///
+/// ## The `_with` Method Suffix
+///
+/// Calling one of the generic `…alloc(x)` methods is essentially equivalent to
+/// the matching [`…alloc_with(|| x)`](?search=alloc_with). However if you use
+/// `…alloc_with`, then the closure will not be invoked until after allocating
+/// space for storing `x` on the heap.
+///
+/// This can be useful in certain edge-cases related to compiler optimizations.
+/// When evaluating for example `bump.alloc(x)`, semantically `x` is first put
+/// on the stack and then moved onto the heap. In some cases, the compiler is
+/// able to optimize this into constructing `x` directly on the heap, however
+/// in many cases it does not.
+///
+/// The `*alloc_with` functions try to help the compiler be smarter. In most
+/// cases doing for example `bump.try_alloc_with(|| x)` on release mode will be
+/// enough to help the compiler realize that this optimization is valid and
+/// to construct `x` directly onto the heap.
+///
+/// ### Warning
+///
+/// These functions critically depend on compiler optimizations to achieve their
+/// desired effect. This means that it is not an effective tool when compiling
+/// without optimizations on.
+///
+/// Even when optimizations are on, these functions do not **guarantee** that
+/// the value is constructed on the heap. To the best of our knowledge no such
+/// guarantee can be made in stable Rust as of 1.44.
+///
+/// ### The `_try_with` Method Suffix
+///
+/// The generic [`…alloc_try_with(|| x)`](?search=_try_with) methods behave
+/// like the purely `_with` suffixed methods explained above. However, they
+/// allow for fallible initialization by accepting a closure that returns a
+/// [`Result`] and will attempt to undo the initial allocation if this closure
+/// returns [`Err`].
+///
+/// #### Warning
+///
+/// If the inner closure returns [`Ok`], space for the entire [`Result`] remains
+/// allocated inside `self`. This can be a problem especially if the [`Err`]
+/// variant is larger, but even otherwise there may be overhead for the
+/// [`Result`]'s discriminant.
+///
+/// <p><details><summary>Undoing the allocation in the <code>Err</code> case
+/// always fails if <code>f</code> successfully made any additional allocations
+/// in <code>self</code>.</summary>
+///
+/// For example, the following will always leak also space for the [`Result`]
+/// into this `Bump`, even though the inner reference isn't kept and the [`Err`]
+/// payload is returned semantically by value:
+///
+/// ```rust
+/// let bump = bumpalo::Bump::new();
+///
+/// let r: Result<&mut [u8; 1000], ()> = bump.alloc_try_with(|| {
+///     let _ = bump.alloc(0_u8);
+///     Err(())
+/// });
+///
+/// assert!(r.is_err());
+/// ```
+///
+///</details></p>
+///
+/// Since [`Err`] payloads are first placed on the heap and then moved to the
+/// stack, `bump.…alloc_try_with(|| x)?` is likely to execute more slowly than
+/// the matching `bump.…alloc(x?)` in case of initialization failure. If this
+/// happens frequently, using the plain un-suffixed method may perform better.
 #[derive(Debug)]
 pub struct Bump {
     // The current chunk we are bump allocating within.
@@ -614,7 +710,7 @@ impl Bump {
     ///
     /// ## Panics
     ///
-    /// Panics if reserving space for `T` would cause an overflow.
+    /// Panics if reserving space for `T` fails.
     ///
     /// ## Example
     ///
@@ -629,38 +725,37 @@ impl Bump {
         self.alloc_with(|| val)
     }
 
+    /// Try to allocate an object in this `Bump` and return an exclusive
+    /// reference to it.
+    ///
+    /// ## Errors
+    ///
+    /// Errors if reserving space for `T` fails.
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// let bump = bumpalo::Bump::new();
+    /// let x = bump.try_alloc("hello");
+    /// assert_eq!(x, Ok(&mut"hello"));
+    /// ```
+    #[inline(always)]
+    #[allow(clippy::mut_from_ref)]
+    pub fn try_alloc<T>(&self, val: T) -> Result<&mut T, alloc::AllocErr> {
+        self.try_alloc_with(|| val)
+    }
+
     /// Pre-allocate space for an object in this `Bump`, initializes it using
     /// the closure, then returns an exclusive reference to it.
     ///
-    /// Calling `bump.alloc(x)` is essentially equivalent to calling
-    /// `bump.alloc_with(|| x)`. However if you use `alloc_with`, then the
-    /// closure will not be invoked until after allocating space for storing
-    /// `x` on the heap.
-    ///
-    /// This can be useful in certain edge-cases related to compiler
-    /// optimizations. When evaluating `bump.alloc(x)`, semantically `x` is
-    /// first put on the stack and then moved onto the heap. In some cases,
-    /// the compiler is able to optimize this into constructing `x` directly
-    /// on the heap, however in many cases it does not.
-    ///
-    /// The function `alloc_with` tries to help the compiler be smarter. In
-    /// most cases doing `bump.alloc_with(|| x)` on release mode will be
-    /// enough to help the compiler to realize this optimization is valid
-    /// and construct `x` directly onto the heap.
-    ///
-    /// ## Warning
-    ///
-    /// This function critically depends on compiler optimizations to achieve
-    /// its desired effect. This means that it is not an effective tool when
-    /// compiling without optimizations on.
-    ///
-    /// Even when optimizations are on, this function does not **guarantee**
-    /// that the value is constructed on the heap. To the best of our
-    /// knowledge no such guarantee can be made in stable Rust as of 1.33.
+    /// See [The `_with` Method Suffix](#the-_with-method-suffix) for a
+    /// discussion on the differences between the `_with` suffixed methods and
+    /// those methods without it, their performance characteristics, and when
+    /// you might or might not choose a `_with` suffixed method.
     ///
     /// ## Panics
     ///
-    /// Panics if reserving space for `T` would cause an overflow.
+    /// Panics if reserving space for `T` fails.
     ///
     /// ## Example
     ///
@@ -703,12 +798,277 @@ impl Bump {
         }
     }
 
+    /// Tries to pre-allocate space for an object in this `Bump`, initializes
+    /// it using the closure, then returns an exclusive reference to it.
+    ///
+    /// See [The `_with` Method Suffix](#the-_with-method-suffix) for a
+    /// discussion on the differences between the `_with` suffixed methods and
+    /// those methods without it, their performance characteristics, and when
+    /// you might or might not choose a `_with` suffixed method.
+    ///
+    /// ## Errors
+    ///
+    /// Errors if reserving space for `T` fails.
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// let bump = bumpalo::Bump::new();
+    /// let x = bump.try_alloc_with(|| "hello");
+    /// assert_eq!(x, Ok(&mut "hello"));
+    /// ```
+    #[inline(always)]
+    #[allow(clippy::mut_from_ref)]
+    pub fn try_alloc_with<F, T>(&self, f: F) -> Result<&mut T, alloc::AllocErr>
+    where
+        F: FnOnce() -> T,
+    {
+        #[inline(always)]
+        unsafe fn inner_writer<T, F>(ptr: *mut T, f: F)
+        where
+            F: FnOnce() -> T,
+        {
+            // This function is translated as:
+            // - allocate space for a T on the stack
+            // - call f() with the return value being put onto this stack space
+            // - memcpy from the stack to the heap
+            //
+            // Ideally we want LLVM to always realize that doing a stack
+            // allocation is unnecessary and optimize the code so it writes
+            // directly into the heap instead. It seems we get it to realize
+            // this most consistently if we put this critical line into it's
+            // own function instead of inlining it into the surrounding code.
+            ptr::write(ptr, f())
+        }
+
+        //SAFETY: Self-contained:
+        // `p` is allocated for `T` and then a `T` is written.
+        let layout = Layout::new::<T>();
+        let p = self.try_alloc_layout(layout)?;
+        let p = p.as_ptr() as *mut T;
+
+        unsafe {
+            inner_writer(p, f);
+            Ok(&mut *p)
+        }
+    }
+
+    /// Pre-allocates space for a [`Result`] in this `Bump`, initializes it using
+    /// the closure, then returns an exclusive reference to its `T` if [`Ok`].
+    ///
+    /// Iff the allocation fails, the closure is not run.
+    ///
+    /// Iff [`Err`], an allocator rewind is *attempted* and the `E` instance is
+    /// moved out of the allocator to be consumed or dropped as normal.
+    ///
+    /// See [The `_with` Method Suffix](#the-_with-method-suffix) for a
+    /// discussion on the differences between the `_with` suffixed methods and
+    /// those methods without it, their performance characteristics, and when
+    /// you might or might not choose a `_with` suffixed method.
+    ///
+    /// For caveats specific to fallible initialization, see
+    /// [The `_try_with` Method Suffix](#the-_try_with-method-suffix).
+    ///
+    /// ## Errors
+    ///
+    /// Iff the allocation succeeds but `f` fails, that error is forwarded by value.
+    ///
+    /// ## Panics
+    ///
+    /// Panics if reserving space for `Result<T, E>` fails.
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// let bump = bumpalo::Bump::new();
+    /// let x = bump.alloc_try_with(|| Ok("hello"))?;
+    /// assert_eq!(*x, "hello");
+    /// # Result::<_, ()>::Ok(())
+    /// ```
+    #[inline(always)]
+    #[allow(clippy::mut_from_ref)]
+    pub fn alloc_try_with<F, T, E>(&self, f: F) -> Result<&mut T, E>
+    where
+        F: FnOnce() -> Result<T, E>,
+    {
+        let rewind_footer = self.current_chunk_footer.get();
+        let rewind_ptr = unsafe { rewind_footer.as_ref() }.ptr.get();
+        let mut inner_result_ptr = NonNull::from(self.alloc_with(f));
+        let inner_result_address = inner_result_ptr.as_ptr() as usize;
+        match unsafe { inner_result_ptr.as_mut() } {
+            Ok(t) => Ok(unsafe {
+                //SAFETY:
+                // The `&mut Result<T, E>` returned by `alloc_with` may be
+                // lifetime-limited by `E`, but the derived `&mut T` still has
+                // the same validity as in `alloc_with` since the error variant
+                // is already ruled out here.
+
+                // We could conditionally truncate the allocation here, but
+                // since it grows backwards, it seems unlikely that we'd get
+                // any more than the `Result`'s discriminant this way, if
+                // anything at all.
+                &mut *(t as *mut _)
+            }),
+            Err(e) => unsafe {
+                // If this result was the last allocation in this arena, we can
+                // reclaim its space. In fact, sometimes we can do even better
+                // than simply calling `dealloc` on the result pointer: we can
+                // reclaim any alignment padding we might have added (which
+                // `dealloc` cannot do) if we didn't allocate a new chunk for
+                // this result.
+                if self.is_last_allocation(NonNull::new_unchecked(inner_result_address as *mut _)) {
+                    let current_footer_p = self.current_chunk_footer.get();
+                    let current_ptr = &current_footer_p.as_ref().ptr;
+                    if current_footer_p == rewind_footer {
+                        // It's still the same chunk, so reset the bump pointer
+                        // to its original value upon entry to this method
+                        // (reclaiming any alignment padding we may have
+                        // added).
+                        current_ptr.set(rewind_ptr);
+                    } else {
+                        // We allocated a new chunk for this result.
+                        //
+                        // We know the result is the only allocation in this
+                        // chunk: Any additional allocations since the start of
+                        // this method could only have happened when running
+                        // the initializer function, which is called *after*
+                        // reserving space for this result. Therefore, since we
+                        // already determined via the check above that this
+                        // result was the last allocation, there must not have
+                        // been any other allocations, and this result is the
+                        // only allocation in this chunk.
+                        //
+                        // Because this is the only allocation in this chunk,
+                        // we can reset the chunk's bump finger to the start of
+                        // the chunk.
+                        current_ptr.set(current_footer_p.as_ref().data);
+                    }
+                }
+                //SAFETY:
+                // As we received `E` semantically by value from `f`, we can
+                // just copy that value here as long as we avoid a double-drop
+                // (which can't happen as any specific references to the `E`'s
+                // data in `self` are destroyed when this function returns).
+                //
+                // The order between this and the deallocation doesn't matter
+                // because `Self: !Sync`.
+                Err(ptr::read(e as *const _))
+            },
+        }
+    }
+
+    /// Tries to pre-allocates space for a [`Result`] in this `Bump`,
+    /// initializes it using the closure, then returns an exclusive reference
+    /// to its `T` if all [`Ok`].
+    ///
+    /// Iff the allocation fails, the closure is not run.
+    ///
+    /// Iff the closure returns [`Err`], an allocator rewind is *attempted* and
+    /// the `E` instance is moved out of the allocator to be consumed or dropped
+    /// as normal.
+    ///
+    /// See [The `_with` Method Suffix](#the-_with-method-suffix) for a
+    /// discussion on the differences between the `_with` suffixed methods and
+    /// those methods without it, their performance characteristics, and when
+    /// you might or might not choose a `_with` suffixed method.
+    ///
+    /// For caveats specific to fallible initialization, see
+    /// [The `_try_with` Method Suffix](#the-_try_with-method-suffix).
+    ///
+    /// ## Errors
+    ///
+    /// Errors with the [`Alloc`](`AllocOrInitError::Alloc`) variant iff
+    /// reserving space for `Result<T, E>` fails.
+    ///
+    /// Iff the allocation succeeds but `f` fails, that error is forwarded by
+    /// value inside the [`Init`](`AllocOrInitError::Init`) variant.
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// let bump = bumpalo::Bump::new();
+    /// let x = bump.try_alloc_try_with(|| Ok("hello"))?;
+    /// assert_eq!(*x, "hello");
+    /// # Result::<_, bumpalo::AllocOrInitError<()>>::Ok(())
+    /// ```
+    #[inline(always)]
+    #[allow(clippy::mut_from_ref)]
+    pub fn try_alloc_try_with<F, T, E>(&self, f: F) -> Result<&mut T, AllocOrInitError<E>>
+    where
+        F: FnOnce() -> Result<T, E>,
+    {
+        let rewind_footer = self.current_chunk_footer.get();
+        let rewind_ptr = unsafe { rewind_footer.as_ref() }.ptr.get();
+        let mut inner_result_ptr = NonNull::from(self.try_alloc_with(f)?);
+        let inner_result_address = inner_result_ptr.as_ptr() as usize;
+        match unsafe { inner_result_ptr.as_mut() } {
+            Ok(t) => Ok(unsafe {
+                //SAFETY:
+                // The `&mut Result<T, E>` returned by `alloc_with` may be
+                // lifetime-limited by `E`, but the derived `&mut T` still has
+                // the same validity as in `alloc_with` since the error variant
+                // is already ruled out here.
+
+                // We could conditionally truncate the allocation here, but
+                // since it grows backwards, it seems unlikely that we'd get
+                // any more than the `Result`'s discriminant this way, if
+                // anything at all.
+                &mut *(t as *mut _)
+            }),
+            Err(e) => unsafe {
+                // If this result was the last allocation in this arena, we can
+                // reclaim its space. In fact, sometimes we can do even better
+                // than simply calling `dealloc` on the result pointer: we can
+                // reclaim any alignment padding we might have added (which
+                // `dealloc` cannot do) if we didn't allocate a new chunk for
+                // this result.
+                if self.is_last_allocation(NonNull::new_unchecked(inner_result_address as *mut _)) {
+                    let current_footer_p = self.current_chunk_footer.get();
+                    let current_ptr = &current_footer_p.as_ref().ptr;
+                    if current_footer_p == rewind_footer {
+                        // It's still the same chunk, so reset the bump pointer
+                        // to its original value upon entry to this method
+                        // (reclaiming any alignment padding we may have
+                        // added).
+                        current_ptr.set(rewind_ptr);
+                    } else {
+                        // We allocated a new chunk for this result.
+                        //
+                        // We know the result is the only allocation in this
+                        // chunk: Any additional allocations since the start of
+                        // this method could only have happened when running
+                        // the initializer function, which is called *after*
+                        // reserving space for this result. Therefore, since we
+                        // already determined via the check above that this
+                        // result was the last allocation, there must not have
+                        // been any other allocations, and this result is the
+                        // only allocation in this chunk.
+                        //
+                        // Because this is the only allocation in this chunk,
+                        // we can reset the chunk's bump finger to the start of
+                        // the chunk.
+                        current_ptr.set(current_footer_p.as_ref().data);
+                    }
+                }
+                //SAFETY:
+                // As we received `E` semantically by value from `f`, we can
+                // just copy that value here as long as we avoid a double-drop
+                // (which can't happen as any specific references to the `E`'s
+                // data in `self` are destroyed when this function returns).
+                //
+                // The order between this and the deallocation doesn't matter
+                // because `Self: !Sync`.
+                Err(AllocOrInitError::Init(ptr::read(e as *const _)))
+            },
+        }
+    }
+
     /// `Copy` a slice into this `Bump` and return an exclusive reference to
     /// the copy.
     ///
     /// ## Panics
     ///
-    /// Panics if reserving space for the slice would cause an overflow.
+    /// Panics if reserving space for the slice fails.
     ///
     /// ## Example
     ///
@@ -737,7 +1097,7 @@ impl Bump {
     ///
     /// ## Panics
     ///
-    /// Panics if reserving space for the slice would cause an overflow.
+    /// Panics if reserving space for the slice fails.
     ///
     /// ## Example
     ///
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -647,6 +647,106 @@ impl Bump {
...
+    ///
+    /// ```
+    /// let bump = bumpalo::Bump::new();
+    /// let x = bump.try_alloc_with(|| Ok("hello"))?;
+    /// assert_eq!(*x, "hello");
+    /// # Result::<_, ()>::Ok(())
+    /// ```
+    #[inline(always)]
+    #[allow(clippy::mut_from_ref)]
+    pub fn try_alloc_with<F, T, E>(&self, f: F) -> Result<&mut T, E>
```
Tamschi, line 703: I'm concerned the naming scheme collides with the methods that error on allocation failure (this one panics), but I couldn't come up with an alternative that still sounded natural.

src/lib.rs
```diff
@@ -647,6 +647,106 @@ impl Bump {
...
+            Ok(t) => Ok(unsafe {
+                //SAFETY:
+                // The `&mut Result<T, E>` returned by `alloc_with` may be
+                // lifetime-limited by `E`, but the derived `&mut T` still has
+                // the same validity as in `alloc_with` since the error variant
+                // is already ruled out here.
+
+                // We could conditionally truncate the allocation here, but since
+                // it grows backwards, it seems unlikely that we'd get any more
+                // than the `Result`'s discriminant this way, if anything at all.
```
Tamschi, line 721: There might be a way to nudge the compiler into end-aligning the payloads in the allocation (so that more space can be freed if the Err payload is larger), but I don't know if that's worth trying at all. There'd also be a branch here, which probably wouldn't be ideal.

tests/try_alloc_with.rs
```diff
@@ -0,0 +1,118 @@
+// All of these try_alloc_with tests will fail with "fatal runtime error: stack overflow" unless
+// LLVM manages to optimize the stack writes away.
+//
+// We only run them when debug_assertions are not set, as we expect them to fail outside release
+// mode.
+
+use bumpalo::Bump;
+
+#[test]
+#[cfg_attr(debug_assertions, ignore)]
```
Tamschi, line 10: If the tests are ignored individually instead of deactivating the entire file, rust-analyzer still shows errors and warnings with default settings.

Tamschi:
Ah whoops, `Debug` isn't available on huge arrays in 1.44. Going to fix that in a moment.

Tamschi:
I'm confident I'll use this now (and certain that the alternative would be either much slower or really messy), so I'm converting this draft into a proper PR.

fitzgen:
Thanks for opening this PR.

Is there a reason that you can't move the fallible parts out of the closure, before the call to `try_alloc_with`? Can you share your example code?

Tamschi:
Sure! For reference, [this is the actual commit in my project](https://github.com/Tamschi/Asteracea/pull/34/commits/c34a19d21c73b4e2cb54be33caccb73b34d733e9). Apologies in advance for the amount of macro templates¹.

The gist is that I'm working on a web frontend framework/templating DSL where the render methods use bumpalo to emit a VDOM graph every frame.² [Individual `Node`s are small](https://github.com/Tamschi/lignin/blob/3973fc7040fda573494c88cecf55482ca8b20d5c/src/lib.rs#L25-L32), so for the most part these are passed around and returned by value.

When it comes to child nodes, I do have to allocate though, so I write the children to the heap like this (de-qualified for brevity):

- `Node::Multi` (VDOM-only grouping) [🔖](https://github.com/Tamschi/Asteracea/pull/34/commits/c34a19d21c73b4e2cb54be33caccb73b34d733e9#diff-f411208a988bea145315eb7f8a98a633e8c28c042667f8ff07e4bdec3c071190L449-R453):
  
  ```rust
  Node::Multi(&*#bump.try_alloc_with(
  	|| -> Result::<_, GUIError> { Ok([
  		#(#m,)*
  	])}
  )?)
  ```

- for HTML elements [🔖](https://github.com/Tamschi/Asteracea/pull/34/commits/c34a19d21c73b4e2cb54be33caccb73b34d733e9#diff-5401d7c2c67900fa7ba493d943897bc343e7bacab33b3a7cf7531c35c784d37eL294-R300):
  
  ```rust
  let children = quote_spanned! {child_stream.span()=>
  	&*#bump.try_alloc_with(
  		|| -> Result<_, GUIError> {
  			Ok([#child_stream])
  		}
  	)?
  };
  ```

Depending on what the individual app does, these arrays could get fairly large, and this would in most cases be hot code, so I'd like to put them onto the heap immediately. The individual items can really be anything, including verbatim Rust blocks and/or expressions, so I have no way of knowing if they're fallible or not outside of very plain HTML expressions.

It's possible I could generate `let` bindings for each element and do the error escalation there, then move everything into the closure (even if that feels icky. I was going for 'as if handwritten' for the generated code, partially to get better compiler errors on type mismatch etc.. Probably not an issue here though).
However, since any error would prevent the allocation in that case, I'm not confident the compiler could reorder that.

These particular `GUIError`s here are essentially panic-equivalent, so once I have the necessary infrastructure otherwise, actually hitting one of them will be very rare and, with the default handlers, eject part of the GUI from further updates. The only reason I haven't made them plain panics is that Rust on Wasm can't unwind yet, but a DI error shouldn't always crash the *entire* application.

- - -

¹ and please disregard where I use `try_alloc_with` for single items, that's [fixed](https://github.com/Tamschi/Asteracea/pull/34/commits/e33463eac1f32abf96bbead54fe9b53702d2db1b) now.

² This was actually partially inspired by Dodrio, but I needed a bit more meta data, so I ended up writing it from scratch. It's not very optimised yet, though, so for now [the DOM renderer](https://github.com/Tamschi/lignin-dom) is 100% Rust.

fitzgen:
Thanks for the context.

As you noted, the `try_` prefix is already used for fallible allocation; we want something different for fallible initialization. Initialization uses the `_with` suffix and I think we could do a `_try_with` suffix for fallible initialization, which is maybe a little clunky but gives us this logical and backwards-compatible table of methods and their fallibility:

| | fallible allocation | infallible allocation |
|---|---|---|
| **fallible init** | `try_alloc_try_with` | `alloc_try_with` |
| **infallible init** | `try_alloc_with` | `alloc_with` |

The `try_alloc_try_with` variant will need a custom error enum that is either an allocation error or the user-supplied closure's error. The `_try_with` variants will also need to free the allocation if initialization fails, rather than leak it.

I would be willing to merge a PR that implements this scheme. I don't think it should be hard to evolve this PR into that.

Tamschi:
Sounds good to me. I'm not sure when, exactly, but I should be able to add the extra methods, tests and benches later this week.

src/lib.rs
```diff
@@ -1128,6 +1269,25 @@ impl Bump {
...
+    // - call f() with the return value being put onto this stack space
+    // - memcpy from the stack to the heap
+    //
+    // Ideally we want LLVM to always realize that doing a stack
+    // allocation is unnecessary and optimize the code so it writes
+    // directly into the heap instead. It seems we get it to realize
+    // this most consistently if we put this critical line into it's
+    // own function instead of inlining it into the surrounding code.
+    ptr::write(ptr, f())
+}
```
Tamschi, line 1289: I moved this function out of of the others to deduplicate it, but I don't think this would make a difference after compiling them. I'd like you to decide where to place them.

tests/try_alloc_with.rs
```diff
@@ -0,0 +1,77 @@
...
+fn try_alloc_with_oom() {
+    todo!()
+}
+
+#[test]
+#[cfg_attr(debug_assertions, ignore)]
+fn try_alloc_with_large_array() {
+    let b = Bump::new();
+
+    b.try_alloc_with(|| [4u8; 10_000_000]).unwrap();
```
Tamschi, line 21: I'd prefer to use `?` here, but `AllocErr` is private, which seems to prevent that.
Is this intentional or should I publish the type as `bumpalo::AllocError`?

src/lib.rs
```diff
@@ -619,34 +619,175 @@ impl Bump {
...
     }
 
+    /// Tries to pre-allocate space for an object in this `Bump`, initializes
+    /// it using the closure, then returns an exclusive reference to it.
+    ///
+    //TODO: Fix or remove the example here. Should there be a try_alloc method?
+    /// Calling `bump.alloc(x)` is essentially equivalent to calling
+    /// `bump.try_alloc_with(|| x)`. However if you use `try_alloc_with`, then the
+    /// closure will not be invoked until after allocating space for storing
+    /// `x` on the heap.
```
Tamschi, line 639: As in the TODO comment, should I add a `try_alloc` method too? I think it would nicely fill a

|                     | fallible allocation  | infallible allocation |
|---------------------|----------------------|-----------------------|
| **by value**        | `try_alloc`          | `alloc`               |
| **infallible init** | `try_alloc_with`     | `alloc_with`          |
| **fallible init**   | `try_alloc_try_with` | `alloc_try_with`      |

pattern, but I'd also understand not wanting to promote the by-value version.

tests/try_alloc_with.rs
```diff
@@ -0,0 +1,77 @@
...
+// mode.
+
+use bumpalo::Bump;
+
+#[test]
+#[cfg_attr(debug_assertions, ignore)]
+#[should_panic(expected = "out of memory")]
+fn try_alloc_with_oom() {
+    todo!()
+}
```
Tamschi, line 14: I'm not too sure I can actually test this. Rust refuses to compile extremely large slices, at least on my PC:

```
error: the type `[u8; 184467440737095516]` is too big for the current architecture
```

Tamschi (commented):
I will stop here for now, but a few design questions came up that I'd need clarification on to complete the PR.

If you have an idea about the OOM test, I'd be curious too. I suspect hitting that code path reliably here isn't exactly feasible, though.

src/lib.rs
Tamschi: Resolved as in https://github.com/fitzgen/bumpalo/pull/91#issuecomment-759079107.

tests/try_alloc_with.rs
Tamschi: I found the implementation in `try_alloc.rs` 😅
I'll see how I can adapt that.

src/lib.rs
```diff
@@ -188,6 +188,24 @@ use core::slice;
 use core::str;
 use core_alloc::alloc::{alloc, dealloc, Layout};
 
+/// An error returned from [`Bump::try_alloc_try_with`].
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum AllocOrInitError<E> {
```
Tamschi, line 193: I went with the slightly more general `AllocOrInitError` over `TryAllocTryWith` here, after all.
I hope exposing the variants is alright, but it's still a purpose-bound error so I *think* this is okay.

The type should also implement `Error` in my opinion, but that would touch on AllocError too, so I'll wait with that for now.

src/lib.rs
```diff
@@ -573,6 +591,27 @@ impl Bump {
         self.alloc_with(|| val)
     }
 
+    //TODO: Tests
```
Tamschi, line 594: I'll try to generalise `try_alloc.rs` a bit, so that the allocator can be shared between different tests.

Tamschi (commented):
Sorry to cause more work for you by asking questions that I end up finding the answers to myself.
Please let them set for a bit. I might be too used to using PRs as TODO-list from work.

I'm going with a "full course" implementation for now since it's not much work and deleting parts of it should be quick and easy.

src/lib.rs
Tamschi: I added all of them for now, since it was next to no effort. Matching tests will follow later this week.

Tamschi:
I believe this completes the PR, unless I made a copy-paste typo somewhere in the documentation and didn't find it yet.
Thanks for considering this!

As an aside, I tried to unify some of the code (see [(try_) unification.patch.txt](https://github.com/fitzgen/bumpalo/files/5824148/try_.unification.patch.txt)), but this made some code paths significantly slower: <details><summary>`cargo bench` log</summary>

**Note:** I'm not sure these measurements are any good, but it's the best I could do on my tablet running only a CMD window and the task manager.

```
    Finished bench [optimized] target(s) in 14.34s
     Running target\release\deps\benches-9e8bb6bc6844d2ce.exe
Gnuplot not found, disabling plotting
alloc/small             time:   [30.787 us 31.114 us 31.545 us]
                        thrpt:  [317.01 Melem/s 321.40 Melem/s 324.81 Melem/s]
                 change:
                        time:   [-3.6026% -1.1280% +1.3719%] (p = 0.39 > 0.05)
                        thrpt:  [-1.3534% +1.1408% +3.7372%]
                        No change in performance detected.
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) high mild
  7 (7.00%) high severe
Benchmarking alloc/big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 15.4s or reduce sample count to 30
alloc/big               time:   [3.0308 ms 3.0357 ms 3.0408 ms]
                        thrpt:  [3.2886 Melem/s 3.2941 Melem/s 3.2995 Melem/s]
                 change:
                        time:   [-0.7714% -0.4196% -0.1099%] (p = 0.02 < 0.05)
                        thrpt:  [+0.1100% +0.4214% +0.7774%]
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe

alloc-with/small        time:   [40.164 us 41.466 us 42.741 us]
                        thrpt:  [233.97 Melem/s 241.16 Melem/s 248.98 Melem/s]
                 change:
                        time:   [+4.7684% +8.4004% +12.367%] (p = 0.00 < 0.05)
                        thrpt:  [-11.006% -7.7494% -4.5514%]
                        Performance has regressed.
Benchmarking alloc-with/big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 15.3s or reduce sample count to 40
alloc-with/big          time:   [3.0190 ms 3.0240 ms 3.0292 ms]
                        thrpt:  [3.3012 Melem/s 3.3069 Melem/s 3.3124 Melem/s]
                 change:
                        time:   [-0.1205% +0.2303% +0.5757%] (p = 0.20 > 0.05)
                        thrpt:  [-0.5724% -0.2298% +0.1206%]
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild
  1 (1.00%) high mild

alloc-try-with/small, small
                        time:   [76.797 us 77.065 us 77.355 us]
                        thrpt:  [129.27 Melem/s 129.76 Melem/s 130.21 Melem/s]
                 change:
                        time:   [+1.5634% +2.0473% +2.5131%] (p = 0.00 < 0.05)
                        thrpt:  [-2.4515% -2.0063% -1.5393%]
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
Benchmarking alloc-try-with/small, big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 35.2s or reduce sample count to 20
alloc-try-with/small, big
                        time:   [6.9628 ms 6.9709 ms 6.9792 ms]
                        thrpt:  [1.4328 Melem/s 1.4345 Melem/s 1.4362 Melem/s]
                 change:
                        time:   [-2.6759% -2.3330% -1.9903%] (p = 0.00 < 0.05)
                        thrpt:  [+2.0307% +2.3887% +2.7495%]
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe
Benchmarking alloc-try-with/big, small: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 32.8s or reduce sample count to 20
alloc-try-with/big, small
                        time:   [6.4648 ms 6.4749 ms 6.4852 ms]
                        thrpt:  [1.5420 Melem/s 1.5444 Melem/s 1.5468 Melem/s]
                 change:
                        time:   [-3.0961% -2.6848% -2.2692%] (p = 0.00 < 0.05)
                        thrpt:  [+2.3219% +2.7589% +3.1951%]
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
Benchmarking alloc-try-with/big, big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 23.6s or reduce sample count to 30
alloc-try-with/big, big time:   [4.6599 ms 4.6679 ms 4.6764 ms]
                        thrpt:  [2.1384 Melem/s 2.1423 Melem/s 2.1460 Melem/s]
                 change:
                        time:   [-0.2901% +0.0719% +0.4742%] (p = 0.74 > 0.05)
                        thrpt:  [-0.4719% -0.0719% +0.2909%]
                        No change in performance detected.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe

alloc-try-with-err/small, small
                        time:   [114.27 us 114.54 us 114.82 us]
                        thrpt:  [87.091 Melem/s 87.306 Melem/s 87.509 Melem/s]
                 change:
                        time:   [-3.7575% -3.2652% -2.7494%] (p = 0.00 < 0.05)
                        thrpt:  [+2.8271% +3.3754% +3.9042%]
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  8 (8.00%) high mild
  3 (3.00%) high severe
Benchmarking alloc-try-with-err/small, big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 23.1s or reduce sample count to 30
alloc-try-with-err/small, big
                        time:   [4.5740 ms 4.5782 ms 4.5828 ms]
                        thrpt:  [2.1821 Melem/s 2.1843 Melem/s 2.1863 Melem/s]
                 change:
                        time:   [+7.2851% +7.5903% +7.8967%] (p = 0.00 < 0.05)
                        thrpt:  [-7.3187% -7.0548% -6.7904%]
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe
Benchmarking alloc-try-with-err/big, small: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 16.5s or reduce sample count to 30
alloc-try-with-err/big, small
                        time:   [3.2458 ms 3.2525 ms 3.2612 ms]
                        thrpt:  [3.0664 Melem/s 3.0745 Melem/s 3.0809 Melem/s]
                 change:
                        time:   [-0.9435% -0.3083% +0.2414%] (p = 0.34 > 0.05)
                        thrpt:  [-0.2409% +0.3093% +0.9525%]
                        No change in performance detected.
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe
Benchmarking alloc-try-with-err/big, big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.8s or reduce sample count to 50
alloc-try-with-err/big, big
                        time:   [1.9423 ms 1.9464 ms 1.9512 ms]
                        thrpt:  [5.1250 Melem/s 5.1376 Melem/s 5.1485 Melem/s]
                 change:
                        time:   [+7.2287% +9.9075% +12.335%] (p = 0.00 < 0.05)
                        thrpt:  [-10.981% -9.0144% -6.7414%]
                        Performance has regressed.

try-alloc/small         time:   [36.296 us 36.894 us 37.618 us]
                        thrpt:  [265.83 Melem/s 271.05 Melem/s 275.52 Melem/s]
                 change:
                        time:   [+1.4426% +4.7548% +8.1826%] (p = 0.00 < 0.05)
                        thrpt:  [-7.5637% -4.5389% -1.4221%]
                        Performance has regressed.
Found 17 outliers among 100 measurements (17.00%)
  2 (2.00%) high mild
  15 (15.00%) high severe
Benchmarking try-alloc/big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 15.5s or reduce sample count to 30
try-alloc/big           time:   [3.0517 ms 3.0568 ms 3.0619 ms]
                        thrpt:  [3.2660 Melem/s 3.2714 Melem/s 3.2768 Melem/s]
                 change:
                        time:   [-0.1380% +0.3655% +0.8382%] (p = 0.14 > 0.05)
                        thrpt:  [-0.8313% -0.3642% +0.1382%]
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe

try-alloc-with/small    time:   [37.099 us 38.511 us 39.941 us]
                        thrpt:  [250.37 Melem/s 259.66 Melem/s 269.55 Melem/s]
                 change:
                        time:   [-6.8818% -3.7387% -0.3998%] (p = 0.03 < 0.05)
                        thrpt:  [+0.4014% +3.8839% +7.3904%]
                        Change within noise threshold.
Found 20 outliers among 100 measurements (20.00%)
  20 (20.00%) high mild
Benchmarking try-alloc-with/big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 15.2s or reduce sample count to 40
try-alloc-with/big      time:   [3.0156 ms 3.0213 ms 3.0267 ms]
                        thrpt:  [3.3039 Melem/s 3.3098 Melem/s 3.3161 Melem/s]
                 change:
                        time:   [+0.0297% +0.3817% +0.7227%] (p = 0.04 < 0.05)
                        thrpt:  [-0.7175% -0.3802% -0.0297%]
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high severe

try-alloc-try-with/small, small
                        time:   [84.321 us 84.470 us 84.637 us]
                        thrpt:  [118.15 Melem/s 118.39 Melem/s 118.59 Melem/s]
                 change:
                        time:   [+5.1242% +5.5777% +6.0258%] (p = 0.00 < 0.05)
                        thrpt:  [-5.6833% -5.2830% -4.8744%]
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe
Benchmarking try-alloc-try-with/small, big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 35.1s or reduce sample count to 20
try-alloc-try-with/small, big
                        time:   [6.9179 ms 6.9297 ms 6.9429 ms]
                        thrpt:  [1.4403 Melem/s 1.4431 Melem/s 1.4455 Melem/s]
                 change:
                        time:   [-3.0434% -2.6458% -2.3006%] (p = 0.00 < 0.05)
                        thrpt:  [+2.3547% +2.7177% +3.1389%]
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
Benchmarking try-alloc-try-with/big, small: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 32.7s or reduce sample count to 20
try-alloc-try-with/big, small
                        time:   [6.4627 ms 6.4693 ms 6.4760 ms]
                        thrpt:  [1.5442 Melem/s 1.5458 Melem/s 1.5473 Melem/s]
                 change:
                        time:   [-3.2540% -2.7620% -2.3663%] (p = 0.00 < 0.05)
                        thrpt:  [+2.4236% +2.8404% +3.3635%]
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  5 (5.00%) high mild
  1 (1.00%) high severe
Benchmarking try-alloc-try-with/big, big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 23.5s or reduce sample count to 30
try-alloc-try-with/big, big
                        time:   [4.6491 ms 4.6557 ms 4.6623 ms]
                        thrpt:  [2.1449 Melem/s 2.1479 Melem/s 2.1510 Melem/s]
                 change:
                        time:   [+1.0471% +1.3128% +1.5711%] (p = 0.00 < 0.05)
                        thrpt:  [-1.5468% -1.2958% -1.0363%]
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild

try-alloc-try-with-err/small, small
                        time:   [119.31 us 119.58 us 119.88 us]
                        thrpt:  [83.413 Melem/s 83.625 Melem/s 83.817 Melem/s]
                 change:
                        time:   [-0.5209% -0.1346% +0.2731%] (p = 0.52 > 0.05)
                        thrpt:  [-0.2724% +0.1347% +0.5237%]
                        No change in performance detected.
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) high mild
  8 (8.00%) high severe
Benchmarking try-alloc-try-with-err/small, big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 22.5s or reduce sample count to 30
try-alloc-try-with-err/small, big
                        time:   [4.4370 ms 4.4419 ms 4.4469 ms]
                        thrpt:  [2.2487 Melem/s 2.2513 Melem/s 2.2538 Melem/s]
                 change:
                        time:   [-0.2756% -0.0040% +0.2875%] (p = 0.98 > 0.05)
                        thrpt:  [-0.2867% +0.0040% +0.2763%]
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high severe
Benchmarking try-alloc-try-with-err/big, small: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 16.4s or reduce sample count to 30
try-alloc-try-with-err/big, small
                        time:   [3.2361 ms 3.2406 ms 3.2455 ms]
                        thrpt:  [3.0811 Melem/s 3.0859 Melem/s 3.0902 Melem/s]
                 change:
                        time:   [-0.3445% -0.0320% +0.2655%] (p = 0.84 > 0.05)
                        thrpt:  [-0.2648% +0.0320% +0.3457%]
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe
Benchmarking try-alloc-try-with-err/big, big: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.2s or reduce sample count to 50
try-alloc-try-with-err/big, big
                        time:   [1.8183 ms 1.8210 ms 1.8238 ms]
                        thrpt:  [5.4830 Melem/s 5.4916 Melem/s 5.4997 Melem/s]
                 change:
                        time:   [-0.5362% -0.1418% +0.2820%] (p = 0.50 > 0.05)
                        thrpt:  [-0.2812% +0.1420% +0.5391%]
                        No change in performance detected.
Found 11 outliers among 100 measurements (11.00%)
  6 (6.00%) high mild
  5 (5.00%) high severe

format-realloc/format-realloc/10
                        time:   [113.18 ns 113.70 ns 114.21 ns]
                        thrpt:  [87.558 Melem/s 87.950 Melem/s 88.357 Melem/s]
                 change:
                        time:   [-0.8027% -0.1835% +0.4548%] (p = 0.57 > 0.05)
                        thrpt:  [-0.4527% +0.1838% +0.8092%]
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
format-realloc/format-realloc/80
                        time:   [113.68 ns 114.24 ns 114.76 ns]
                        thrpt:  [697.12 Melem/s 700.30 Melem/s 703.70 Melem/s]
                 change:
                        time:   [-1.4246% -0.6593% +0.1217%] (p = 0.10 > 0.05)
                        thrpt:  [-0.1216% +0.6637% +1.4452%]
                        No change in performance detected.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
format-realloc/format-realloc/270
                        time:   [113.88 ns 114.38 ns 114.88 ns]
                        thrpt:  [2.3503 Gelem/s 2.3606 Gelem/s 2.3710 Gelem/s]
                 change:
                        time:   [+0.0874% +0.6710% +1.2582%] (p = 0.03 < 0.05)
                        thrpt:  [-1.2425% -0.6665% -0.0874%]
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
format-realloc/format-realloc/640
                        time:   [113.75 ns 114.26 ns 114.72 ns]
                        thrpt:  [5.5787 Gelem/s 5.6014 Gelem/s 5.6262 Gelem/s]
                 change:
                        time:   [-0.6360% -0.0776% +0.5180%] (p = 0.79 > 0.05)
                        thrpt:  [-0.5153% +0.0776% +0.6400%]
                        No change in performance detected.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

</details>

Tamschi:
Well, almost. The question of `alloc::AllocErr` is still open, which also means that `AllocOrInitError<_>` isn't `Error` yet. I can already implement `Display`, though.

src/lib.rs
fitzgen: >  I hope exposing the variants is alright, but it's still a purpose-bound error so I _think_ this is okay.

Yeah this is fine.

> The type should also implement `Error` in my opinion, but that would touch on AllocError too, so I'll wait with that for now.

This crate isn't `std` so we can't implement the `Error` trait, as it isn't available in `core`.

tests/try_alloc_with.rs
fitzgen: It isn't intentional, but I'd like to wait on #92 before poking at this any further.

fitzgen (commented):
Thanks @Tamschi. I left a couple comments inline below, but overall this looks good and we can merge it when those comments are addressed. Can you also squash this into a single commit? Thanks.

tests/try_alloc_with.rs
Tamschi: I'll leave this as-is despite the #92 merge, since it's probably better to clean this up separately instead of burying it in a larger squash commit.

src/lib.rs
```diff
@@ -647,6 +694,285 @@ impl Bump {
...
+                if current_ptr.get() == ptr.cast() {
+                    // We can also reuse the memory, unless `f` made any
+                    // further allocations in `self`.
+                    if current_footer_p == rewind_footer {
+                        current_ptr.set(rewind_ptr)
+                    } else {
+                        // If the current chunk changed, we can at least reset
+                        // to its start, since we know no other allocations
+                        // happened.
+                        current_ptr.set(current_footer_p.as_ref().data)
```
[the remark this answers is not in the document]
Tamschi: > For example, it could have done something like
> 
> ```rust
> let x: Option<&mut u32> = None;
> let _ = b.alloc_try_with(|| {
>     x = Some(b.alloc(42));
>     Err(())
> });
> x.unwrap()
> ```

This will make the `current_ptr.get() == ptr.cast()` check fail, regardless of the chunk change. I've clarified the comments regarding that.

(Please bear with me while I go through these one by one. I'll squash and rebase everything afterwards.)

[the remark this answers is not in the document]
Tamschi: > It seems like all this hairy unsafe pointer code could be removed by reusing the `<Bump as Alloc>::dealloc` method here, which already makes sure it only actually adjusts the bump pointer if the given pointer points to the last allocation made.

The reason I didn't use `dealloc` is that that method won't free alignment bytes, even though we have the necessary `rewind_ptr` here to do this.

If leaking them is okay, I'll change it, but I personally would prefer to do this as cleanly (in that regard) as possible.

tests/try_alloc_with.rs
Tamschi: I think so, or at least to file a note about it with rust-analyzer. It would be nice to be able to specify multiple check configurations, per crate in the workspace.
Iterating all the cfgs that can't change dependencies would probably cause a lot of (tooling system requirements) overhead for projects that don't need this, though.

I'd personally still keep this here since using `ignore` has the nice side effect of highlighting the tests in the `cargo test` output if they didn't run, but I'd use such a configuration for a few other projects.

src/lib.rs
```diff
@@ -647,6 +694,285 @@ impl Bump {
...
+    #[allow(clippy::mut_from_ref)]
+    pub fn alloc_try_with<F, T, E>(&self, f: F) -> Result<&mut T, E>
+    where
+        F: FnOnce() -> Result<T, E>,
+        E: Unpin,
+    {
+        let rewind_footer = self.current_chunk_footer.get();
+        let rewind_ptr = unsafe { rewind_footer.as_ref() }.ptr.get();
+        let ptr = NonNull::from(self.alloc_with(f));
+        match unsafe { ptr.clone().as_mut() } {
```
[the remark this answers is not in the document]
Tamschi: It's a borrowing/mutability issue (in both places):
Without the `.clone()`, `.as_mut()` doesn't compile as it takes `&mut self`. Conversely, if I make `ptr` `mut` instead, then that's an exclusive borrow and I can't do the `ptr` comparison in the `Err(…)` branch.

I'll rewrite this a bit though, to make it clear dereferencing that address is UB in the match.

src/lib.rs
```diff
@@ -573,6 +600,26 @@ impl Bump {
         self.alloc_with(|| val)
     }
 
+    /// Try to allocate an object in this `Bump` and return an exclusive
+    /// reference to it.
+    ///
+    /// ## Errors
+    ///
+    /// Errors if reserving space for `T` would cause an overflow.
```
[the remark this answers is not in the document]
Tamschi: Whoops, thanks for the catch!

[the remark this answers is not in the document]
Tamschi: This happens in many places, also on existing methods, so I was so free to update it to "if reserving space … fails" everywhere.
I think the pointer overflow is an implementation detail and doesn't need to be mentioned explicitly, but correct me if I'm wrong.

Tamschi:
@fitzgen I think I've addressed everything now, aside from your concerns about not using `<Bump as Alloc>::dealloc` (see my other comment).

I also made a few smaller last-minute changes to bits and pieces I noticed while reading through the full changeset again, like removing the `Unpin` bounds. You can find those separately by checking the commits on [my squash backup branch](https://github.com/Tamschi/bumpalo/commits/backup/try_alloc_with).

src/lib.rs
```diff
@@ -703,12 +798,233 @@ impl Bump {
...
+                    if current_footer_p == rewind_footer {
+                        current_ptr.set(rewind_ptr)
+                    } else {
+                        // If the current chunk changed, we can at least reset
+                        // to its start, since we know no other allocations
+                        // happened (because `current_ptr` still matches
+                        // `inner_result_address`).
+                        current_ptr.set(current_footer_p.as_ref().data)
+                    }
+                }
```
fitzgen, line 931: Okay, I see why we would like to avoid `dealloc`, so we can reclaim any alignment padding added for this allocation (thanks for explaining!) However, I'd still like to reuse the `is_last_allocation` helper so that some of these pointer comparisons aren't open coded and we use "semantic" methods that make the code read smoother and better clarify intent. Additionally, I rewrote the comments "proving" soundness to convince myself that everything was above board here, and (maybe this is biased, since I wrote it, but) I think these new commetns read better:

```rust
    // If this result was the last allocation in this arena, we can reclaim its
    // space. In fact, sometimes we can do even better than simply calling
    // `dealloc` on the result pointer: we can reclaim any alignment padding we
    // might have added (which `dealloc` cannot do) if we didn't allocate a new
    // chunk for this result.
    if self.is_last_allocation(inner_result_address) {
        if current_footer_p == rewind_footer {
            // It's still the same chunk, so reset the bump pointer to its
            // original value upon entry to this method (reclaiming any
            // alignment padding we may have added).
            current_ptr.set(rewind_ptr);
        } else {
            // We allocated a new chunk for this result.
            //
            // We know the result is the only allocation in this chunk: Any
            // additional allocations since the start of this method could only
            // have happened when running the initializer function, which is
            // called *after* reserving space for this result. Therefore, since
            // we already determined via the check above that this result was
            // the last allocation, there must not have been any other
            // allocations, and this result is the only allocation in this
            // chunk.
            //
            // Because this is the only allocation in this chunk, we can reset
            // the chunk's bump finger to the start of the chunk.
            current_ptr.set(current_footer_p.as_ref().data);
        }
    }
```

src/lib.rs
```diff
@@ -703,12 +798,233 @@ impl Bump {
...
+            }),
+            Err(e) => unsafe {
+                //SAFETY:
+                // As we received `E` semantically by value from `f`, we can just
+                // copy that value here as long as we avoid a double-drop (which
+                // can't happen as any specific references to the `E`'s data in
+                // `self` are destroyed when this function returns).
+                let current_footer_p = self.current_chunk_footer.get();
+                let current_ptr = &current_footer_p.as_ref().ptr;
+                if current_ptr.get().as_ptr() as usize == inner_result_address {
```
fitzgen, line 1003: Same thing here.

tests/try_alloc.rs
```diff
@@ -91,6 +91,36 @@ fn main() {
...
+        // Unlike with `try_alloc_layout`, it's not that easy to test a variety
+        // of size/capacity combinations here.
+        // Since nothing in Bump is really random, and we have to start fresh
+        // each time, just checking each case once is enough.
+        for &fail_alloc in &[false, true] {
+            let bump = GLOBAL_ALLOCATOR.with_successful_allocs(|| {
+                // We can't query the remaining space in the current chunk,
+                // so we have to create a new Bump for each test and fill it to
+                // the brink of a new allocation.
+                let bump = Bump::try_new().unwrap();
```
fitzgen, line 104: Sorry I missed this before, but the `chunk_capacity` method gives this information, so we shouldn't need to create a whole new bump every time.

fitzgen (commented):
Okay, final few nitpicks and then we can merge this. Thanks for your patience @Tamschi!

tests/try_alloc.rs
Tamschi: I think there might be slightly ambiguous wording between my comment and the documentation here:

```rust
    /// Gets the remaining capacity in the current chunk (in bytes).
    ///
    /// […]
    pub fn chunk_capacity(&self) -> usize {
        let current_footer = self.current_chunk_footer.get();
        let current_footer = unsafe { current_footer.as_ref() };

        current_footer as *const _ as usize - current_footer.data.as_ptr() as usize
    }
```

`chunk_capacity` returns the total capacity that remains after allocating the header (as expected by its name, in my opinion), but I'd need the live value of remaining unallocated space to do this reliably.

It might still be possible to work with that when making assumptions about chunk/`ChunkFooter` and `u8` alignments, but that doesn't feel like a clean solution...

Since the code is hit only eight times during testing, I'd like to leave it as-is for simplicity's sake.
Actually exposing the remaining free space in a chunk wouldn't be useful, I think, since it can't be used to predict whether a given layout would allocate.

I'll adjust the comment to make it unambiguous, though.

src/lib.rs
Tamschi: Thank you, this is better than what I wrote. I'll reflow the comments for 80 character lines (+ add a necessary cast), but will otherwise use it verbatim.

fitzgen:
Looks like there is a valgrind failure in CI that wasn't in the previous iteration of this PR. Mind looking into it?

As soon as that is resolved and CI is green, I'll merge this! Thanks @Tamschi!

Tamschi:
I saw that, but it looked like these were in tests that don't access any code this pull request touches.
From what I can tell, it's in the panic handling that happens here:

```rust
#[test]
#[should_panic(expected = "out of memory")]
fn alloc_slice_overflow() {
    let b = Bump::new();

    b.alloc_slice_fill_default::<u64>(usize::max_value());
}
```

Still, I'll do my best to figure out what's going on.

Tamschi:
I believe this is a Valgrind bug.

Valgrind 3.13.0, which passed on master, [reports](https://github.com/fitzgen/bumpalo/runs/1752417127#step:7:264):

```
--8686-- WARNING: unhandled amd64-linux syscall: 332
--8686-- You may be able to write your own handler.
--8686-- Read the file README_MISSING_SYSCALL_OR_IOCTL.
--8686-- Nevertheless we consider this a bug.  Please report
--8686-- it at http://valgrind.org/support/bug_reports.html.
```

Valgrind 3.15.0, which fails here, [reports](https://github.com/fitzgen/bumpalo/runs/1785071066#step:7:269):

```
==7387== Syscall param statx(file_name) points to unaddressable byte(s)
==7387==    at 0x4B0989D: syscall (syscall.S:38)
==7387==    by 0x18849B: statx (weak.rs:95)
[…]
```

The Valgrind 3.16.0 changelog (found [here](https://www.valgrind.org/docs/manual/dist.news.html)) says:
> * ==================== FIXED BUGS ====================
> […]
> n-i-bz  sys_statx: don't complain if both |filename| and |buf| are NULL.
> […]

Valgrind 3.14.0 (same page):
> * ==================== FIXED BUGS ====================
> […]
> 392118  unhandled amd64-linux syscall: 332 (statx)
> […]

fitzgen:
Merged in #94, along with the updated valgrind suppressions. Thanks again @Tamschi!

Tamschi:
Thank you too, @fitzgen, for the help and criticism, and of course for accepting the feature request!

I indirectly mentioned this before, but succeeding with this PR gives me some piece of mind regarding where I'm going with my project's implementation details, so I'm really glad it happened. I hope the added functions will come in handy for others too, of course.

PR #92, opened 2021-01-16
Implement allocator_api

### LINKED ISSUE #87: Provide feature to use nightly allocator API
In latest Rust nightly we can allocate with custom allocator (for example https://doc.rust-lang.org/nightly/alloc/vec/struct.Vec.html#method.new_in).

It will be great for bumpalo to provide the ability to use this. To achieve this, the `&Bump` needs to implement `std::alloc::AllocRef`.

baloo:
supersedes #68: different approach, collection implementation is kept and this only implements the api needed to satisfy `alloc::alloc::Allocator`
fixes #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,8 @@ Eventually [all `std` collection types will be parameterized by an
 allocator](https://github.com/rust-lang/rust/issues/42774) and we can remove
 this `collections` module and use the `std` versions.
 
+see `allocator_api` support below on nightly.
+
 ### `bumpalo::boxed::Box`
 
 When the `"boxed"` cargo feature is enabled, a fork of `std::boxed::Box` library
@@ -149,6 +151,10 @@ example in `rayon`.
 
 The [`bumpalo-herd`](https://crates.io/crates/bumpalo-herd) crate provides a pool of `Bump`
 allocators for use in such situations.
+
+### `allocator_api` support
+
+Bumpalo provides nightly' `allocator_api` support via the cargo feature `allocator_api`.
 
 #### Minimum Supported Rust Version (MSRV)
 
--- a/README.md
+++ b/README.md
@@ -98,7 +98,8 @@ Eventually [all `std` collection types will be parameterized by an
 allocator](https://github.com/rust-lang/rust/issues/42774) and we can remove
 this `collections` module and use the `std` versions.
 
-see `allocator_api` support below on nightly.
+For unstable, nightly-only support for custom allocators in `std`, see the
+`allocator_api` section below.
 
 ### `bumpalo::boxed::Box`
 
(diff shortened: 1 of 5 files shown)


baloo:
cc @NOBLES5E @hansihe

Amanieu:
rust-lang/hashbrown#231 is currently blocked on this to support bumpalo for hashmaps.

README.md
```diff
@@ -98,6 +98,8 @@ Eventually [all `std` collection types will be parameterized by an
 allocator](https://github.com/rust-lang/rust/issues/42774) and we can remove
 this `collections` module and use the `std` versions.
 
+see `allocator_api` support below on nightly.
```
fitzgen, line 101:
```suggestion
For unstable, nightly-only support for custom allocators in `std`, see the
`allocator_api` section below.
```

README.md
```diff
@@ -150,6 +152,10 @@ example in `rayon`.
 The [`bumpalo-herd`](https://crates.io/crates/bumpalo-herd) crate provides a pool of `Bump`
 allocators for use in such situations.
 
+### `allocator_api` support
+
+Bumpalo provides nightly' `allocator_api` support via the cargo feature `allocator_api`.
```
fitzgen, line 157: Can you add more docs here briefly describing what the `allocator_api` nightly feature enables and provide a complete example that shows

* turning on the feature
* creating a `Bump`
* creates a `Vec` backed by the bump via the `new_in` method

Thanks

src/lib.rs
```diff
@@ -168,6 +174,10 @@ to only do so it when done deliberately and for good reasons.
 #![deny(missing_debug_implementations)]
 #![deny(missing_docs)]
 #![no_std]
+#![cfg_attr(
+    feature = "allocator_api",
+    feature(allocator_api, nonnull_slice_from_raw_parts)
```
fitzgen, line 179: Is the `nonnull_slice_from_raw_parts` feature necessary in order to implement the `Allocator` API? If not, I'd like to avoid using it, and only depend on `allocator_api`. This will expose the crate to less nightly breakage and make it that much easier to go stable when the time comes.

src/lib.rs
```diff
@@ -1247,6 +1259,24 @@ unsafe impl<'a> alloc::Alloc for &'a Bump {
     }
 }
 
+#[cfg(feature = "allocator_api")]
+unsafe impl<'a> Allocator for &'a Bump {
```
fitzgen, line 1263: Can you also implement `grow` and `shrink`? See the existing `realloc` method for details (and maybe make `realloc` branch and call these new methods, so there isn't code duplication)

src/lib.rs
```diff
@@ -1247,6 +1259,24 @@ unsafe impl<'a> alloc::Alloc for &'a Bump {
...
+            .map_err(|_| AllocError)
+    }
+
+    unsafe fn deallocate(&self, ptr: NonNull<u8>, layout: Layout) {
+        // If the pointer is the last allocation we made, we can reuse the bytes,
+        // otherwise they are simply leaked -- at least until somebody calls reset().
+        if self.is_last_allocation(ptr) {
+            let ptr = NonNull::new_unchecked(ptr.as_ptr().add(layout.size()));
+            self.current_chunk_footer.get().as_ref().ptr.set(ptr);
+        }
```
fitzgen, line 1276: Can this either call `<Bump as GlobalAlloc>::dealloc` or vice versa? I'd rather not have the pointer manipulation and all that duplicated in multiple places.

fitzgen (commented):
Thanks! A few inline comments to address, and then we can merge this.

src/lib.rs
baloo: Done

src/lib.rs
baloo: I believe that would require changing the implementation of `try_alloc_layout_fast` and `try_alloc_layout_slow`. Which ultimately will have to happen once you drop the Box and Vec implementation, but... If you have any other suggestion, I'm all ears, since this is scary business :D

fitzgen: I guess it is fine then, if the `allocator_api` feature stabilizes before `nonnull_slice_from_raw_parts` (I kinda doubt it) then we can cross this bridge at that time.

fitzgen (approved):
Thanks!

baloo:
Thank you!

fitzgen:
Published 3.5.0 with this change.

Issue #98, opened 2021-02-17
allocator_api: missing grow_zeroed implementation


baloo:
I believe the implementation made in https://github.com/fitzgen/bumpalo/pull/92 could use a specific `grow_zeroed` implementation

If I read https://github.com/rust-lang/rust/blob/ee88f46bb5e27c4d9f30326e69ff2298dcbeecb1/library/core/src/alloc/mod.rs#L253-L277 correctly, that could use more effective one.
As currently implemented, it grows by `new_layout.size()` where it could grow by only `new_layout.size()-old_layout.size()`.

PR #99, opened 2021-02-17
allocator_api: provide a grow_zeroed implementation

baloo:
Fixup #98

The current implementation of grow_zeroed in liballoc allocates a new chunk, copies data over, then deallocates the old chunk.
Bumpalo can use a more memory efficient version of that, and this is an implementation of it.


baloo:
(there is some noise due to rustfmt, which I forgot to run in the initial implementation, let me know if you want me to drop it).

fitzgen (approved):
Thanks!

PR #114, opened 2021-06-03
Add reset_back

quininer:
This adds another reset method, which is different from reset in that it keeps last chunk on linked list instead of first chunk.

This is helpful for keep expected memory size after reset.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -765,6 +765,61 @@ impl Bump {
         }
     }
 
+    /// Reset this bump allocator.
+    ///
+    /// Like `reset`, but keep the last chunk instead of the first chunk.
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// let mut bump = bumpalo::Bump::new();
+    ///
+    /// // Allocate a bunch of things.
+    /// {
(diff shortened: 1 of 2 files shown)


konsumlamm:
I don't see the point of this new method. What advantage does it have over `reset`? What do you mean by "keep expected memory size"?

DJankauskas:
I'm assuming what this does is instead of deallocating all previously allocated chunks other than the first one, it keeps all allocations, allowing you to reuse them when you reuse `Bump`. This seems useful to me, since if you're performing a bunch of tasks with similar memory needs, it might make sense to prevent having to fetch chunks from the global allocator over and over again. Would this not be so performant in practice? If not, this PR seems like something I might use in a project I'm working on right now (though I'd have to benchmark it).

Chaoses-Ib:
This can be used to reserve capacity. Currently, one can only only reserve capacity at the start (by `with_capacity` or `alloc` + `reset`). With `reset_back`, one can reserve needed space by `alloc` + `reset_back` in the middle.

fitzgen:
This is already the existing behavior of `reset`.

Issue #100, opened 2021-03-07
Allow creation of allocator from slice


vinaychandra:
The current `new` methods use Global allocator for the allocator to work with. Provide an `unsafe` way to create a bump allocator which will allow to create the allocator from a slice denoting unused memory.

fitzgen:
It is a pretty fundamental assumption in this implementation that chunks of memory come from the global allocator, that we can request new chunks as needed, and that we deallocate the chunks when the `Bump` itself is deallocated.

Changing this assumption would require either dynamic checks or making the whole crate parameterized by another allocator (which would dole out portions of a given slice for your requested feature). The first option is unacceptable. The second is possible, but would require some care and thought, and I'm not totally sure would be worth it or not. I also don't have the cycles to explore the second option right now.

The easiest thing to do for your use case would probably be to write a specialized crate for exactly what you're trying to do (possibly forking/borrowing code from this crate).

PR #115, opened 2021-06-16
Add custom allocator support for Bump

EliaGeretto:
This pull requests adds support for specifying a custom allocator that `Bump` should use to allocate new chunks. This should provide a reasonable interface for situations in which memory should not be retrieved from the `Global` allocator and swapping the whole allocator using `#[global_allocator]` is not possible.

I have taken care to make this compile both with nightly and stable Rust, albeit some nasty generics workarounds were necessary.

This PR should, ideally, solve #100.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -243,9 +243,11 @@ use core::mem;
 use core::ptr::{self, NonNull};
 use core::slice;
 use core::str;
-use core_alloc::alloc::{alloc, dealloc, Layout};
+use core_alloc::alloc::Layout;
+#[cfg(not(feature = "allocator_api"))]
+use core_alloc::alloc::{alloc, dealloc};
 #[cfg(feature = "allocator_api")]
-use core_alloc::alloc::{AllocError, Allocator};
+use core_alloc::alloc::{AllocError, Allocator, Global};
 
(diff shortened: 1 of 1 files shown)


EliaGeretto:
Could you please tell me if you would be willing to merge this and, in that case, provide some comments?

fitzgen:
Thanks for making this pull request and giving us an idea of what changes are required to support generic backing allocators in `bumpalo`.

> Could you please tell me if you would be willing to merge this and, in that case, provide some comments?

Unlike implementing the allocator API, this support for being parameterized by an allocator is rather invasive with all those `cfg`s. Therefore, I'd prefer to wait until the allocator API stabilizes -- and `bumpalo` can always be parameterized by an allocator, rather than only being parameterized in some build configurations -- before adding this support.

EliaGeretto:
I can understand. I need this for my own project, so I am planning to keep my branch reasonably up-to-date. I will leave the PR here in case someone else needs this functionality.

EliaGeretto:
I am not sure why, but to compile correctly my code requires at least Rust 1.48.

fitzgen:
Closing old pull requests.

Issue #116, opened 2021-08-11
Continuous address allocated


sundy-li:
Hello, can this crate support Continuous address allocated? 
The function returns the first allocated reference. And we can use offset to access b and c.
``` 
pub fn alloc_with<F,  A, B, C>(&self, a: A, b: B, c: B) -> &mut A
```

fitzgen:
You can use a `#[repr(C)]` struct to do this:

```rust
#[repr(C)]
struct Abc {
    a: A,
    b: B,
    c: C,
}

let bump = Bump::new();
let abc = bump.alloc_with(|| Abc { a: foo(), b: bar(), c: baz() });
do_stuff(&mut abc.a);
```

I don't think this is a common enough usage pattern to warrant adding sugar to `bumpalo` itself, however.

PR #117, opened 2021-08-11
add alloc_with_inplace

sundy-li:
This may help in #116

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -788,7 +788,7 @@ impl Bump {
     /// Simiar to `alloc`, but rewrite the memory in place without any allocation.
     #[inline(always)]
     #[allow(clippy::mut_from_ref)]
-    pub fn alloc_in_place<T>(&self, place: NonNull<u8>, val: T) -> &mut T {
+    pub unsafe fn alloc_in_place<T>(&self, place: NonNull<u8>, val: T) -> &mut T {
         self.alloc_with_inplace(place, || val)
     }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -858,6 +865,28 @@ impl Bump {
         }
     }
 
+    /// Similar to alloc_with, but rewrite the memory in place without any allocation.
+    #[inline(always)]
+    #[allow(clippy::mut_from_ref)]
+    pub fn alloc_with_inplace<F, T>(&self, place: NonNull<u8>, f: F) -> &mut T
```
fitzgen, line 871: Neither of these functions can be safe because there is no guarantee that the `NonNull<u8>` is a valid memory location to store a `T`, let alone previously allocated by this `Bump`.

fitzgen (commented):
It isn't clear to me that this functionality is necessary or even more ergonomic than using `std::ptr::write` directly. Can you make the case for why this should be included, what the motivation is, etc.

sundy-li:
> It isn't clear to me that this functionality is necessary or even more ergonomic than using std::ptr::write directly. Can you make the case for why this should be included, what the motivation is, etc.

Ok, If we want to allocate states in the database system. Like query:

`select age, count(score), max(score), min(score) from table group by age`.

We will use a hashmap to store the aggregate function states addresses for future merging, eg: `countState`, `maxState`, `minState`.

Like: `map<key,  vec<usize> >`.   Among it `vec<usize>` represents the addresses.  If the dataset is very large, we can improve it using `map<Key, usize>`, among `usize` is the address of `countState` object.

But `bumpalo` is using LinkedList chunk to store the memory address, so the address may not be continuous even we know the layout of each object.

`Allocate A` ---> `Allocate B` ---> `Allocate C`  ,  `A`, `B` and `C` may not in the some Chunk. So we know the address of `A`, but it's not available to know the address of `B` using address offset if we did not store it.

So let's allocate a large memory firstly, which the layout can hold `sum( sizeof(A), sizeof(B), sizeof(c))` by using `allocate_layout`, then we can initial `A,B,C` in the specific addresses.

PR #118, opened 2021-08-28
Fix docs, Bump is !Sync, not !Send

untitaker:


untitaker:
cc @vorner who added this paragraph

vorner:
> cc @vorner who added this paragraph

I guess I just wasn't paying enough attention and confused the two. The docs and compiler seems to agree with you, not with me ;-)

untitaker:
thanks it's just I have no idea what I am doing

fitzgen (approved):
Good catch, thanks!

fitzgen:
The valgrind failure is the `cargo readme`-is-up-to-date test failing. @untitaker can you run `cargo readme -t README.tpl > README.md` so that the `README.md` is synced with the module level docs? That should fix this failure.

The rust 1.44 CI failure looks like `cargo readme` doesn't build on rust 1.44 anymore, and is unrelated to this crate actually building on rust 1.44. I think we can either:

* use the latest Rust to install `cargo readme` and then build with our rust 1.44, or
* add an env var or something for skipping the `cargo readme` test, and skip it in the rust 1.44 builder.

I don't have time to do this latter fix myself at the moment, so if you have time @untitaker that would be super appreciated. Or anyone else!

untitaker:
Do you want to keep the `cargo readme` test in the rust testsuite at all costs? I was thinking that this could be its own CI-job which auto-pushes the readme update for PRs

fitzgen:
`cargo-readme` fix is over in https://github.com/fitzgen/bumpalo/pull/122

untitaker:
sorry for dropping the ball on this, rebased PR!

fitzgen (approved):
Thanks!

PR #120, opened 2021-09-05
Reduce crate size

lrazovic:
Using the `include` field in the `Cargo.toml` 23% (or 100.9 KB) can be saved in 18 files (of 433.6 KB and 33 files in entire crate).
The final `.crate` file goes from 131KB to 76.5 KB.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ name = "bumpalo"
 readme = "./README.md"
 repository = "https://github.com/fitzgen/bumpalo"
 version = "3.7.0"
+include = ["src/**/*", "LICENSE-*", "README.*", "CHANGELOG.md"]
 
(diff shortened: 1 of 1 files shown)


fitzgen:
I'd prefer to use `exclude` rather than `include`, just so that it is less likely that I forget to include some file or directory in the future that is now required because of some recent change and then break the whole package anyone depending on the crate. I'd prefer the failure mode where I forget to add new `exclude`s and the packages gets a tiny bit more bloated than it should be, rather than breaking things for users.

Happy to merge if you flip this around into an `exclude`!

lrazovic:
Flipped from `include` to `exclude`, with the same size decrease. Now from the .crate are excluded tests, benchmarks and especially the logo.

fitzgen (approved):
Thanks!

PR #125, opened 2021-10-15
Add CollectIn, FromIteratorIn traits to allow for easier construction

dreamsmasher:
Currently, building `Bump`-parameterized collections from iterators is a little verbose, e.g.

```rust
use Bumpalo{Bump, collections::Vec};
let bump = Bump::new();
let xs =  Vec::from_iter_in((0..10), &bump);
```
while a normal `Vec` can be constructed directly using `.collect()`. 

This PR adds 2 new traits, `FromIteratorIn`, and an extension trait `CollectIn` that allows for simpler constructor of bump-parameterized structures. Currently, `Vec` and `String` implement `FromIteratorIn`, and future collections should try to do the same if possible.

On iterators, `CollectIn` allows a more chainable syntax for constructing these collections:

```rust
let bump = Bump::new();
let xs = (0..10).collect_in::<Vec<i32>>(&bump);
```

Quickcheck-based tests are also included and passing.

### DIFF
--- a/tests/collect_in.rs
+++ b/tests/collect_in.rs
@@ -1,0 +1,29 @@
+#![cfg(feature = "collections")]
+use bumpalo::collections::{CollectIn,Vec, String};
+use bumpalo::Bump;
+use quickcheck::{quickcheck};
+use std::string::String as StdString;
+use std::vec::Vec as StdVec;
+
+
+#[cfg(test)]
+quickcheck! {
+  fn test_string_collect(input: StdString) -> bool {
+    let bump = Bump::new();
+    let bump_str = input.chars().collect_in::<String>(&bump);
+    &bump_str == &input
+  }
+}
+
+#[cfg(test)]
+quickcheck! {
+  fn test_vec_collect(input: StdVec<i32>) -> bool {
+    let bump = Bump::new();
+    let bump_vec = input.clone().into_iter().collect_in::<Vec<_>>(&bump);
+    let bump_ref: &[i32] = &bump_vec;
+
+    bump_ref == &input
+  }
+
+
+}
--- a/src/collections/collect_in.rs
+++ b/src/collections/collect_in.rs
@@ -1,0 +1,54 @@
+use super::{Vec, String};
+use crate::Bump;
+
+/// Wrapper trait for types that support being constructed from an iterator, parameterized by an allocator.
+pub trait FromIteratorIn<A> {
+    /// An allocator type
+    type Alloc;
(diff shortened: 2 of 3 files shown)


tests/collect_in.rs
```diff
@@ -0,0 +1,29 @@
...
+  fn test_vec_collect(input: StdVec<i32>) -> bool {
+    let bump = Bump::new();
+    let bump_vec = input.clone().into_iter().collect_in::<Vec<_>>(&bump);
+    let bump_ref: &[i32] = &bump_vec;
+
+    bump_ref == &input
+  }
+
+
+}
```
fitzgen, line 29: Can you make sure there is a newline at the end of the file? I think running `cargo fmt` will do this.

fitzgen:
> Would you like things to be moved to a `collections_traits` module still?

No, this PR is fine, since it isn't modifying the files that are borrowed from `std`. Thanks!

fitzgen:
https://github.com/fitzgen/bumpalo/pull/125/checks?check_run_id=3929433788#step:6:80

```
error[E0277]: can't compare `[i32]` with `std::vec::Vec<i32>`
  --> tests/collect_in.rs:25:14
   |
25 |     bump_ref == &input
   |              ^^ no implementation for `[i32] == std::vec::Vec<i32>`
   |
   = help: the trait `std::cmp::PartialEq<std::vec::Vec<i32>>` is not implemented for `[i32]`
   = note: required because of the requirements on the impl of `std::cmp::PartialEq<&std::vec::Vec<i32>>` for `&[i32]`
```

src/collections/vec.rs
```diff
@@ -1896,6 +1896,16 @@ impl<'bump, T: 'bump> Extend<T> for Vec<'bump, T> {
...
+impl<'bump, T> FromIteratorIn<T> for Vec<'bump, T> {
+    type Alloc = &'bump Bump;
+    fn from_iter_in<I>(iter: I, alloc: Self::Alloc) -> Self
+    where
+        I: IntoIterator<Item = T>,
+    {
+        Vec::from_iter_in(iter, alloc)
+    }
+}
+
```
fitzgen, line 1908: Can you avoid modifying the files that are copied from `std`? This makes it easier to update these files in the future.

dreamsmasher: Oh, I didn't realize these were straight from `std`. I'll move the impls back to their original place, sorry!

dreamsmasher: Oh, I didn't realize these were straight from `std`. I'll move the impls back to their original place, sorry!

PR #126, opened 2021-10-15
Add `iter_allocated_chunks_unchecked()`

### LINKED ISSUE #121: Allow iter_allocated_chunks on a &self
In [starlark-rust](https://github.com/facebookexperimental/starlark-rust) we use Bumpalo as the backing for a garbage collector. Starlark has the property that after a module has been interpreted, its heap is frozen, which leads us to have `struct FrozenHeap(Arc<Bumpalo>)`. We'd like to add a feature to iterate over the heap to produce memory profile reports, but alas, `iter_allocated_chunks` takes a `&mut self`, which we can't produce given an immutable `Bumpalo` in an `Arc`. Would it be possible to provide an `iter_allocated_chunks` variant which took a non-`mut` `self`, perhaps in `unsafe` if necessary. While I'm happy to call `unsafe`, converting my `&Bumpalo` into an `&mut Bumpalo` given many owners is instant undefined behaviour in Rust.

poconn:
Fixes #121

As per discussion in the issue, we want (unsafe) iteration over an arena's chunks without a mutable borrow of `self`.

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1594,8 +1594,9 @@ impl Bump {
     ///
     /// ## Safety
     ///
-    /// This arena must not be allocated from while the returned iterator is
-    /// alive.
+    /// Must not be called if there currently exist any mutable references to
+    /// data allocated from this arena. This arena must not be allocated from
+    /// while the returned iterator is alive.
     ///
     /// In addition, all of the caveats when reading the chunk data from
     /// [`iter_allocated_chunks()`](Bump::iter_allocated_chunks) still apply.


src/lib.rs
```diff
@@ -1579,7 +1579,27 @@ impl Bump {
     ///     assert_eq!(chunk[2].assume_init(), b'a');
     /// }
     /// ```
+    #[inline(always)]
     pub fn iter_allocated_chunks(&mut self) -> ChunkIter<'_> {
+        // SAFE: `self` is borrowed mutably.
+        unsafe { self.iter_allocated_chunks_unchecked() }
```
poconn, line 1585: Removed a little bit of newly-duplicated code, but I can revert this section if you want to avoid a new `unsafe` usage (even if trivial) for not much benefit.

src/lib.rs
```diff
@@ -1579,7 +1579,27 @@ impl Bump {
...
+    /// responsible for ensuring that the arena is not modified while iterating.
+    ///
+    /// ## Safety
+    ///
+    /// This arena must not be allocated from while the returned iterator is
+    /// alive.
+    ///
+    /// In addition, all of the caveats when reading the chunk data from
+    /// [`iter_allocated_chunks()`](Bump::iter_allocated_chunks) still apply.
+    pub unsafe fn iter_allocated_chunks_unchecked(&self) -> ChunkIter<'_> {
```
poconn, line 1602: Not sure about the name, "unchecked" sound to me like it omits asserts or something which isn't accurate, open to suggestions.

poconn:
Just noticed this part of @fitzgen's comment on the issue which I missed earlier:

> `iter_allocated_chunks` gives our shared references to the underlying chunks which, if we allowed `&self` on `iter_allocated_chunks`, could have active exclusive references from allocated things

Need to mention that in the doc also, will post a revision.

poconn:
Sure, makes a lot of sense, thanks! Posted a new PR: https://github.com/fitzgen/bumpalo/pull/127

PR #127, opened 2021-10-18
Add `iter_allocated_chunks_raw()`

### LINKED ISSUE #121: Allow iter_allocated_chunks on a &self

poconn:
Adds `iter_allocated_chunks_raw()`, which (a) yields raw `(*mut u8, usize)`, and (b) only borrows the arena immutably.

Fixes #121, follow-up from PR https://github.com/fitzgen/bumpalo/pull/126 (posting a new one as all of the previous changes would be reverted)

I need to finish the docs and add tests, posting just to get confirmation on this approach first.

Thanks!

### DIFF
--- a/tests/quickchecks.rs
+++ b/tests/quickchecks.rs
@@ -108,7 +108,7 @@ fn overlap((a1, a2): (usize, usize), (b1, b2): (usize, usize)) -> bool {
 fn contains((a1, a2): (usize, usize), (b1, b2): (usize, usize)) -> bool {
     assert!(a1 < a2);
     assert!(b1 < b2);
-    a1 <= b1 && a2 >= b2
+    a1 <= b1 && b2 <= a2
 }
 
 fn range<T>(t: &T) -> (usize, usize) {
@@ -255,10 +255,9 @@ quickcheck! {
         }
         let raw_chunks: Vec<(_, _)> = unsafe { b.iter_allocated_chunks_raw() }.collect();
         let chunks: Vec<&[_]> = b.iter_allocated_chunks().collect();
+        assert_eq!(raw_chunks.len(), chunks.len());
         for ((ptr, size), chunk) in raw_chunks.into_iter().zip(chunks) {
-            if let Some(byte) = chunk.get(0) {
-                assert_eq!(ptr as *const u8, byte.as_ptr());
-            }
+            assert_eq!(ptr as *const _, chunk.as_ptr() as *const _);
             assert_eq!(size, chunk.len());
         }
     }


src/lib.rs
```diff
@@ -1724,12 +1755,7 @@ impl<'a> Iterator for ChunkIter<'a> {
         unsafe {
             let foot = self.footer?;
             let foot = foot.as_ref();
-            let data = foot.data.as_ptr() as usize;
-            let ptr = foot.ptr.get().as_ptr() as usize;
-            debug_assert!(data <= ptr);
-            debug_assert!(ptr <= foot as *const _ as usize);
-
-            let len = foot as *const _ as usize - ptr;
+            let (ptr, len) = foot.as_raw_parts();
```
fitzgen, line 1758: `ChunkIter` could contain a `raw: ChunkRawIter` inside of it and then this impl could just be

```rust
let (ptr, len) = self.raw.next()?;
let slice = slice::from_raw_parts(ptr as *const mem::MaybeUninit<u8>, len);
Some(slice)
```

instead of duplicating the footer-walking logic.

fitzgen (commented):
Yeah this looks good.

Can you add some smoke tests for iteration as well? Maybe a quickcheck that allocates a bunch of things, records the ranges where the things are allocated, and then makes sure that every allocated thing is within one of the chunks as returned by `iter_allocated_chunks_raw`? See `tests/quickchecks.rs` for the other quickchecks, which should give you an idea of how to write this kind of test.

Thanks!

tests/quickchecks.rs
```diff
@@ -227,4 +234,32 @@ quickcheck! {
...
+
+    fn chunks_and_raw_chunks_are_same(values: Vec<BigValue>) -> () {
+        let mut b = Bump::new();
+        for val in values {
+            b.alloc(val);
+        }
+        let raw_chunks: Vec<(_, _)> = unsafe { b.iter_allocated_chunks_raw() }.collect();
+        let chunks: Vec<&[_]> = b.iter_allocated_chunks().collect();
+        for ((ptr, size), chunk) in raw_chunks.into_iter().zip(chunks) {
+            if let Some(byte) = chunk.get(0) {
```
poconn, line 259: If we want to check the address of empty chunks, seems like there isn't a technically correct way to do it in stable Rust, [this nightly method](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.slice_as_ptr) is what we need.

In practice, `chunk.as_ptr() as *const u8` should work (using the slice's `.as_ptr()` instead of `MaybeUninit`, assuming that the address of the `MaybeUninit` is the same as the address of the contained value) - which is true for `std`'s current implementation and "should" always be, but AFAICT not actually guaranteed. Maybe that's an acceptable amount of unspecified behavior (not actual UB) to rely on in a test case?

tests/quickchecks.rs
```diff
@@ -104,6 +104,13 @@ fn overlap((a1, a2): (usize, usize), (b1, b2): (usize, usize)) -> bool {
     a1 < b2 && b1 < a2
 }
 
+// Returns whether `(b1, b2)` is contained in `(a1, a2)`.
+fn contains((a1, a2): (usize, usize), (b1, b2): (usize, usize)) -> bool {
+    assert!(a1 < a2);
+    assert!(b1 < b2);
+    a1 <= b1 && a2 >= b2
```
fitzgen, line 111: nitpick, but I like it this way because you can visually see that `b` is within `a`:

```suggestion
    a1 <= b1 && b2 <= a2
```

tests/quickchecks.rs
```diff
@@ -227,4 +234,32 @@ quickcheck! {
...
+        }
+    }
+
+    fn chunks_and_raw_chunks_are_same(values: Vec<BigValue>) -> () {
+        let mut b = Bump::new();
+        for val in values {
+            b.alloc(val);
+        }
+        let raw_chunks: Vec<(_, _)> = unsafe { b.iter_allocated_chunks_raw() }.collect();
+        let chunks: Vec<&[_]> = b.iter_allocated_chunks().collect();
```
fitzgen, line 257: Can you also add `assert_eq!(chunks.len(), raw_chunks.len())` here? `zip` will silently ignore elements from one of the iterators if they aren't both the same size.

fitzgen (approved):
Tests look great!

Couple nitpicks below and then this should be ready to merge.

Thanks!

Issue #133, opened 2021-11-23
Vec<'bump, T> does not call Drop for its items


wada314:
Hello, I'm not sure if this is an intentional design decision or just a bug though, `Vec<'bump, T>` does not call the owning items' `Drop` when itself is dropped.
I'll send a PR to add the `Drop` impl for `Vec` later so please utilize it if it's not an intended behavior, thanks! 😃 

```rust
#[cfg(test)]
use bumpalo::{boxed::Box, collections::Vec, Bump};
use std::cell::RefCell;
#[cfg(test)]
use std::ops::Deref;

struct Foo<'a>(&'a RefCell<String>);
impl<'a> Drop for Foo<'a> {
    fn drop(&mut self) {
        self.0.borrow_mut().push_str("Dropped!");
    }
}

#[test]
fn test_bare() {
    let buffer = RefCell::new(String::new());
    {
        let _bare_foo = Foo(&buffer);
    }
    // OK
    assert_eq!("Dropped!", buffer.borrow().deref());
}

#[test]
fn test_boxed() {
    let buffer = RefCell::new(String::new());
    let bump = Bump::new();
    {
        let _boxed_foo = Box::new_in(Foo(&buffer), &bump);
    }
    // OK
    assert_eq!("Dropped!", buffer.borrow().deref());
}

#[test]
fn test_vec() {
    let buffer = RefCell::new(String::new());
    let bump = Bump::new();
    {
        let mut vec_foo = Vec::new_in(&bump);
        vec_foo.push(Foo(&buffer));
    }
    // Fails!!
    assert_eq!("Dropped!", buffer.borrow().deref());
}
```

wada314:
Thank you for merging it! 🙇

PR #134, opened 2021-11-23
Add impl Drop for Vec<'bump, T>

wada314:
For fixing #133.
The lib code is just a copy-and-paste from the official [vec/mod.rs](https://github.com/rust-lang/rust/blob/2e055d92e0d527b273d12584bd842f6527e7652c/library/alloc/src/vec/mod.rs#L2813).


fitzgen (approved):
Thanks!

PR #140, opened 2022-01-10
Improve documentation

konsumlamm:
* improve doc comments
* fix & add links
* add & improve examples

PR #141, opened 2022-01-10
Fix CI

konsumlamm:
CI broke, since you apparently renamed `master` to `main` (I wondered why it didn't run on [my PR](https://github.com/fitzgen/bumpalo/pull/140)). I also took the liberty to rename `rust.yml` to `rust.yaml`, as `.yaml` is the [officially recommended file extension](https://yaml.org/faq.html).


fitzgen (approved):
Thanks, good catch!

Issue #143, opened 2022-02-15
Question: Is it okay for Bump::grow to only receive a new_size, but not the entire new_layout?


yanchith:
Was doing some relaxing reading of bumpalo's source (commit https://github.com/fitzgen/bumpalo/commit/b22416cfae42ad3c13814c5c97364725196f07ab) when I found the following:

`Allocator::grow` receives `new_layout`, but the Allocator trait impl in bumpalo only passes `new_size` to `Bump::grow`.

What happens, if `Allocator::grow` gets called with a new layout that has a greater alignment requirement? Docs of `Allocator::grow` say:

> Returns Err if the new layout does not meet the allocator’s size and alignment constraints of the allocator, or if growing otherwise fails.

So I guess it would be ok to return `AllocError` if the alignment requirements got stricter with the new layout, but the trait impl currently ignores that.

I also could be missing something 😂

fitzgen:
@yanchith thanks for filing an issue! Yes, I think that `Bump::grow` should take a new layout so that it can handle the new alignment requirements (if any) as well.

Would you be up for making these changes and sending a pull request?

yanchith:
If you mean just the lazy version of returning `AllocError`, if the new alignment isn't automatically satisfied, then sure! I think this might also apply to `Bump::shrink`, so I'll look into that as well.  I'll be afk for the rest of the week, but I think I could send the PR afterwards.

Btw are you privy to what lead to the decision of letting the caller request a different alignment when growing/shrinking in the allocator api?

fitzgen:
> If you mean just the lazy version of returning `AllocError`, if the new alignment isn't automatically satisfied, then sure!

That's a good place to start, but it shouldn't be hard to add the more general support (just call `alloc` again).

> Btw are you privy to what lead to the decision of letting the caller request a different alignment when growing/shrinking in the allocator api?

I am not. Might be worth bringing up with the `wg-allocators` folks, if you feel like going down the rabbit hole.

yanchith:
FYI I asked around and got pointed here regarding the design history: https://github.com/rust-lang/wg-allocators/issues/5

yanchith:
I think we can now close this - feel free reopen if you want this for documentation.

PR #144, opened 2022-02-21
Account for alignment changes when growing or shrinking

yanchith:
My riff on #143

Both `Bump::grow` and `Bump::shrink` now take in the new layout and check for alignment compatibility. `Bump::grow` falls back to `Bump::try_alloc_layout` if the alignment is not compatible, and `Bump::shrink` just returns `AllocError` in this case.

These above strategies are pretty much up to us. The Allocator trait says it is ok to error out, if it can't handle the new requirements. For bump allocation, I'd personally prefer shrinking to fail instead of allocating, if the alignment requirements changed.

Afaik, the collections in alloc don't change alignment during their life (unless transmuted), and while I can imagine legitimate code that would want to do this, the same code would probably write their own allocator.

Also added two tests, but `allocator_grow_layout_change` doesn't really test anything just yet. It is a bit tricky to figure out when grow falls back to `Bump::try_alloc_layout` without instrumenting Bump itself with counters. Ideas?

### DIFF
--- a/tests/allocator_api.rs
+++ b/tests/allocator_api.rs
@@ -1,6 +1,7 @@
-#![feature(allocator_api, slice_ptr_get)]
+#![feature(allocator_api)]
 #![cfg(feature = "allocator_api")]
 use bumpalo::Bump;
+use quickcheck::quickcheck;
 
 use std::alloc::{AllocError, Allocator, Layout};
 use std::ptr::NonNull;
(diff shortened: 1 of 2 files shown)


fitzgen:
> These above strategies are pretty much up to us. The Allocator trait says it is ok to error out, if it can't handle the new requirements. For bump allocation, I'd personally prefer shrinking to fail instead of allocating, if the alignment requirements changed.

Agreed, this sounds like the correct trade off to me.

tests/allocator_api.rs
```diff
@@ -1,4 +1,4 @@
-#![feature(allocator_api)]
+#![feature(allocator_api, slice_ptr_get)]
```
fitzgen, line 1: Could you avoid the use of new nightly features here? It just makes things harder when we have to update unrelated tests because the feature has changed in some breaking way since the last PR was made. I know that the tests will be slightly more verbose, but it is worth it to avoid forcing churn upon unrelated PRs.

tests/allocator_api.rs
```diff
@@ -114,3 +114,31 @@ fn allocator_grow_zeroed() {
...
+    let layout_align4 = Layout::from_size_align(1024, 4).unwrap();
+    let layout_align16 = Layout::from_size_align(256, 16).unwrap();
+
+    // Allocate a chunk of memory and attempt to shrink it while increasing
+    // alignment requirements.
+    let p4 = b.allocate(layout_align4).unwrap().as_non_null_ptr();
+    let p16_res = unsafe { b.shrink(p4, layout_align4, layout_align16) };
+
+    assert_eq!(p16_res, Err(AllocError));
+}
```
fitzgen, line 144: Regarding being unsure whether the tests exercise the fallback-to-new-alloc paths, I think we could have much more confidence in exercising this code if we wrote targeted quickcheck tests instead of unit tests for these APIs.

I'm imagining that we would do something like:

```rust
quickcheck! {
    fn allocator_grow_layout_change(layouts: Vec<(usize, usize)>) -> () {
        let mut layouts = layouts.into_iter().map(|(size, align)| {
            let align = std::cmp::min(align, 64);
            let align = round_up_to_pow2(align);
            let size = std::cmp::min(size, 1024);
            let size = std::cmp::max(size, align);
            Layout::from_size_align(size, align).unwrap()
        }).collect::<Vec<_>>();
        layouts.sort_by_key(|l| (l.size(), l.align()));
        // TODO: allocate the first layout in the bump and then grow
        // to each of the next layouts
    }
}
```

where `round_up_to_pow2` is basically this helper but without checked arithmetic (safe because we are clamping the input just above): https://cs.github.com/bytecodealliance/wasmtime/blob/b438617e1238bec01ebc804c0087e4849df0dc87/crates/runtime/src/externref.rs?q=round_up#L291

Exercising shrinking would be the exact same except that we would invert the comparator so that the layouts were ordered form largest to smallest.

Does that make sense?

fitzgen (commented):
Thanks! This is looking great. Just some suggestions for how to better exercise these paths in tests.

tests/allocator_api.rs
yanchith: Yes, this makes sense. Do we want to assert something in these quickchecks, or is it enough that they don't panic for now?

fitzgen: I think we can start with just asserting that the grown/shrunk pointer has the requested alignment.

yanchith:
Sooo, I added those quickchecks, and actually found what looks like a real issue in `Bump::shrink`: applying the size delta to the pointer could actually produce pointers with insufficient alignment (because the size could actually be any "unround" integer once it is GTE the alignment). I fixed this by aligning the delta, so doing arithmetic with it doesn't change alignment of the pointer. We reclaim less space this way, but the new pointer stays correctly aligned.

This is all very new to me, so please double-check 😂

fitzgen (commented):
Thanks! I think the `shrink` code was assuming that the size would always be a multiple of the alignment but it is just greater-than-or-equal. Good find!

Just one nitpick to fix below and then we can merge this.

fitzgen (approved):
Thanks! And thanks for sticking with this PR throughout a couple rounds of feedback!

PR #145, opened 2022-04-18
Remove ptr-int-ptr casts

saethlin:
This replaces ptr-int-ptr roundtrips as well as ptr-int casts which were
used to produce usize values to do arithmetic on instead of doing
pointer arithmetic operations on pointers to u8. These ptr-int-ptr
roundtrips are a thorny case for aliasing models to deal with, and thus
play havoc with unsafe checkers such as Miri, and complicate alias
analysis and aliasing models. Removing them does not necessarily remove
UB, but it significantly simplifies the semantics of this code and makes
it easier to rule out definite forms of UB.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -277,12 +277,12 @@ impl ChunkFooter {
     // Returns the start and length of the currently allocated region of this
     // chunk.
     fn as_raw_parts(&self) -> (*const u8, usize) {
-        let data = self.data.as_ptr() as usize;
-        let ptr = self.ptr.get().as_ptr() as usize;
+        let data = self.data.as_ptr() as *const u8;
+        let ptr = self.ptr.get().as_ptr() as *const u8;
         debug_assert!(data <= ptr);
-        debug_assert!(ptr <= self as *const _ as usize);
-        let len = self as *const _ as usize - ptr;
-        (ptr as *const u8, len)
+        debug_assert!(ptr <= self as *const ChunkFooter as *const u8);
+        let len = unsafe { (self as *const ChunkFooter as *const u8).offset_from(ptr) as usize };
+        (ptr, len)
     }
 }
 
@@ -488,9 +488,9 @@ impl Bump {
             let data = NonNull::new(data)?;
 
             // The `ChunkFooter` is at the end of the chunk.
-            let footer_ptr = data.as_ptr() as usize + new_size_without_footer;
+            let footer_ptr = data.as_ptr().add(new_size_without_footer);
             debug_assert_eq!((data.as_ptr() as usize) % align, 0);
-            debug_assert_eq!(footer_ptr % CHUNK_ALIGN, 0);
+            debug_assert_eq!(footer_ptr as usize % CHUNK_ALIGN, 0);
             let footer_ptr = footer_ptr as *mut ChunkFooter;
 
             // The bump pointer is initialized to the end of the range we will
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1213,13 +1213,18 @@ impl Bump {
...
-            let start = footer.data.as_ptr() as usize;
+            let ptr = footer.ptr.get().as_ptr();
+            let start = footer.data.as_ptr();
             debug_assert!(start <= ptr);
-            debug_assert!(ptr <= footer as *const _ as usize);
+            debug_assert!(ptr as *const u8 <= footer as *const _ as *const u8);
 
-            let ptr = ptr.checked_sub(layout.size())?;
-            let aligned_ptr = ptr & !(layout.align() - 1);
+            if (ptr as usize) < layout.size() {
```
fitzgen, line 1221: Isn't this one of the pointer-to-int casts that this PR is trying to avoid?

src/lib.rs
```diff
@@ -1213,13 +1213,18 @@ impl Bump {
...
+                return None;
+            }
+
+            let ptr = ptr.wrapping_sub(layout.size());
+            let rem = ptr as usize % layout.align();
```
fitzgen, line 1226: Similar here?

tests/quickchecks.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg(not(miri))]
```
fitzgen, line 1: Can you add a comment describing why these tests are disabled when running under miri?

Nitpick: newline between this attribute and the `use` statements, please.

```suggestion
#![cfg(not(miri))]

```

tests/tests.rs
```diff
@@ -7,8 +7,12 @@ use std::usize;
 fn can_iterate_over_allocated_things() {
     let mut bump = Bump::new();
 
+    #[cfg(not(miri))]
     const MAX: u64 = 131_072;
 
+    #[cfg(miri)]
+    const MAX: u64 = 1024;
```
fitzgen, line 14: Similarly, can you add a comment here explaining why `MAX` is smaller when running under miri? (I assume we time out / take too long to run tests because running under miri is slow?)

tests/tests.rs
```diff
@@ -154,10 +159,13 @@ where
 #[test]
 fn with_capacity_test() {
     with_capacity_helper(0u8..255);
-    with_capacity_helper(0u16..10000);
-    with_capacity_helper(0u32..10000);
-    with_capacity_helper(0u64..10000);
-    with_capacity_helper(0u128..10000);
+    #[cfg(not(miri))]
```
fitzgen, line 162: Same here.

fitzgen (commented):
Thanks for making this PR. I've wanted to get miri running in CI for a while but haven't had the cycles to get it set up myself.

I am a little confused as to when it is still okay to convert a pointer to an int, as this PR doesn't remove all of them / introduces new such conversions. See inline comments below.

Other than that, I just have a couple nitpicks.

Thanks!

saethlin:
This PR is not concerned with ptr-int casts, it is concerned with ptr-int-ptr roundtrips done via casts. The problem with these roundtrips is that selecting a provenance for the new pointer is a nightmare. As far as I can tell, a formal aliasing model which supports provenance-based optimizations (which are done in all major compilers) needs to pick at least one of:
1. Ban `restrict`/`noalias` from the language (rustc already emits `noalias` so this would be a bit odd)
2. Pick a model that has (exponential?) state space explosion/memory usage inside a checking interpreter
3. Rely on a very complicated formal model which nobody has yet come up with
4. Make dereferencing a pointer that has been round-tripped through an integer UB

It seems like C is going for option 2 and Rust is going with one of 3 or 4. Option 4 is the only option which is actionable for users of Rust code, and I'm pursuing it because it's the only way we have to run a checker for an aliasing model which accomodates the optimizations produced by LLVM's `noalias` attribute, which rustc currently applies to mutable references. This is `MIRIFLAGS=-Zmiri-tag-raw-pointers`. By default Miri checks a more complicated aliasing model which permits integers to be cast back to pointers, but it permits code which is UB with how rustc currently emits `noalias`.

The Strict Provenance experiment is concerned with eliminating ptr-int casts. It would be nifty if we could do this, but doing so is not an option for library implementors right now because the whole idea relies on a bunch of standard library APIs that are behind a feature gate, and the API you're supposed to use to inspect the bits of a pointer is currently implemented with a ptr-int cast with a TODO comment about making it an intrinsic.

---

Alignment checks in particular suck. A lot. We should eventually have a `#![feature(strict_provenance)]` API for doing alignment checks. Currently, `align_offset` exists but (and it documents this clearly) it is _completely broken_ in `const fn` and perhaps in Miri (because Miri mostly uses the `const fn` machinery).

fitzgen:
Ah okay, thanks. I was misunderstanding that it is specifically the last link of `pointer --> int --> pointer` that this PR was eliminating. Everything makes sense now.

Happy to merge once the nitpicks are addressed.

fitzgen (approved):
Thanks!

PR #148, opened 2022-04-29
Fix Vec::push in Miri tests with -Zmiri-tag-raw-pointers

overdrivenpotato:
This avoids going through the slice deref, fixing a Miri error. It might be worth porting over the current `std::vec` implementation, as the current one seems to be old and the error doesn't occur in `std::vec` anymore. But this is just a quick band-aid fix for this particular error.


fitzgen:
> It might be worth porting over the current `std::vec` implementation

I certainly don't have the energy for this, but if you do, I'm happy to receive a PR!

PR #150, opened 2022-05-03
Expose `drain_filter`

### LINKED ISSUE #137: `drain_filter` is private
The `Vec::drain_filter` method is private, eventhough the `DrainFilter` struct it returns is public (and it tries to link to `Vec::drain_filter`, which fails, because it is private). I assume this is an oversight? It is also missing documentation.

konsumlamm:
Closes #137.


fitzgen (approved):
Thanks!

PR #151, opened 2022-05-03
Export `AllocErr`

### LINKED ISSUE #139: `alloc::AllocErr` is not exported
`AllocErr` itself is marked public, but the module it's in (`alloc`) isn't, so it isn't exported. However, `AllocErr` is used in a lot of places, e.g. `AllocOrInitError::Alloc` and `Bump::try_alloc`. This also breaks a bunch of links in the documentation.

`AllocErr` is just a unit struct that has no information attached, but it's still weird having a type in a pubic API that is not exported.

konsumlamm:
Resolves #139.


fitzgen (approved):
Thanks!

PR #153, opened 2022-05-18
The scope of the unsafe block can be appropriately reduced

peamaeq:
In this function you use the unsafe keyword for almost the entrie function body. However, I found that only 6 functions and 1 operation are real unsafe operations (see the list below). 

We need to mark unsafe operations more precisely using unsafe keyword. Keeping unsafe blocks small can bring many benefits. For example, when mistakes happen, we can locate any errors related to memory safety  within an unsafe block. This is the balance between Safe and Unsafe Rust. The separation is designed to make using Safe Rust as ergonomic as possible, but requires extra effort and care when writing Unsafe Rust. 
**Real unsafe operation list:**
1. the layout_from_size_align()/alloc()/new_unchecked()/write()/as_ref()/sub() function(these are unsafe functions)
2. dereferencing raw pointers(&mut *p)

Hope this PR can help you.
Best regards.
**References**
https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html 
https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -446,7 +446,7 @@ impl Bump {
         requested_layout: Option<Layout>,
         prev: Option<NonNull<ChunkFooter>>,
     ) -> Option<NonNull<ChunkFooter>> {
-        unsafe {
+        
             let mut new_size_without_footer =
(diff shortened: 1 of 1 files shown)


fitzgen:
Closing old pull requests.

PR #160, opened 2022-06-02
Add support for allocation limits on arenas

Eliasin:
Implements #135 

Arenas can optionally keep track of an `allocation_limit` which is
considered when allocating new chunks to an arena.

The limit is only checked in `alloc_layout_slow` and `new_chunk`, which means it
should have a low performance overhead as the limit is only checked when trying to hit
the global allocator.

To further avoid performance concerns, `ChunkFooter` has been refactored to keep
track of the total allocated bytes, allowing `Bump::allocated_bytes` to avoid
performing a walk of the chunk list.

No new `new` methods were added, and limits must be set after the
creation of a `Bump` arena. Limits can be changed on the fly and are
only enforced when attempting to allocate new chunks.

No changes were made to the errors returned by operations like `alloc`
because of backwards compatibility.

I'm new to open source contributions and library development so sorry if there's anything obvious I missed.

### DIFF
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -99,7 +99,7 @@ fn size_align<T>() -> (usize, usize) {
 }
 
 /// The `AllocErr` error indicates an allocation failure
-/// that may be due to resource exhaustion, reaching an allocation limit, or
+/// that may be due to resource exhaustion or to
 /// something wrong when combining the given input arguments with this
 /// allocator.
 // #[unstable(feature = "allocator_api", issue = "32838")]
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -99,7 +99,7 @@ fn size_align<T>() -> (usize, usize) {
 }
 
 /// The `AllocErr` error indicates an allocation failure
-/// that may be due to resource exhaustion or to
+/// that may be due to resource exhaustion, reaching an allocation limit, or
 /// something wrong when combining the given input arguments with this
 /// allocator.
 // #[unstable(feature = "allocator_api", issue = "32838")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -476,6 +476,12 @@ fn allocation_size_overflow<T>() -> T {
     panic!("requested allocation size overflowed")
 }
 
+// This can be migrated to directly use `usize::abs_diff` when the MSRV
+// reaches `1.60`
+fn abs_diff(a: usize, b: usize) -> usize {
+    usize::max(a, b) - usize::min(a, b)
+}
+
 impl Bump {
     /// Construct a new arena to bump allocate into.
     ///
@@ -594,7 +600,7 @@ impl Bump {
             if allocated_bytes > allocation_limit {
                 None
             } else {
-                Some(allocation_limit.abs_diff(allocated_bytes))
+                Some(abs_diff(allocation_limit, allocated_bytes))
             }
         })
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -452,6 +452,8 @@ const FIRST_ALLOCATION_GOAL: usize = 1 << 9;
 // take the alignment into account.
 const DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER: usize = FIRST_ALLOCATION_GOAL - OVERHEAD;
 
+/// The memory size and alignment details for a potential new chunk
+/// allocation.
 #[derive(Debug, Clone, Copy)]
 struct NewChunkMemoryDetails {
     new_size_without_footer: usize,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -255,8 +255,8 @@ impl<E: Display> Display for AllocOrInitError<E> {
 ///
 /// ### `Bump` Allocation Limits
 ///
-/// `Bump` supports setting a limit on the maximum bytes of memory that can
-/// be allocated for use. This limit can be set and removed with
+/// `bumpalo` supports setting a limit on the maximum bytes of memory that can
+/// be allocated for use in a particular `Bump` arena. This limit can be set and removed with
 /// [`set_allocation_limit`][Bump::set_allocation_limit] and
 /// [`remove_allocation_limit`][Bump::remove_allocation_limit].
 /// Changing the limit for a `Bump` while it has live allocations does
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -258,9 +258,9 @@ impl<E: Display> Display for AllocOrInitError<E> {
 /// `bumpalo` supports setting a limit on the maximum bytes of memory that can
 /// be allocated for use in a particular `Bump` arena. This limit can be set and removed with
 /// [`set_allocation_limit`][Bump::set_allocation_limit].
-/// Changing the limit for a `Bump` while it has live allocations does
-/// nothing until a new allocation is attempted, at which point it may fail
-/// due to the new limit.
+/// The allocation limit is only enforced when allocating new backing chunks for
+/// a `Bump`. Updating the allocation limit will not affect existing allocations
+/// or any future allocations within the `Bump`'s current chunk.
 ///
 /// ## Example
 ///
@@ -557,9 +557,9 @@ impl Bump {
 
     /// Set the allocation limit in bytes for this arena.
     ///
-    /// Changing the limit for a `Bump` while it has live allocations does
-    /// nothing until a new allocation is attempted, at which point it may fail
-    /// due to the new limit.
+    /// The allocation limit is only enforced when allocating new backing chunks for
+    /// a `Bump`. Updating the allocation limit will not affect existing allocations
+    /// or any future allocations within the `Bump`'s current chunk.
     ///
     /// ## Example
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -587,17 +587,14 @@ impl Bump {
     /// How much headroom an arena has before it hits its allocation
     /// limit.
     fn allocation_limit_remaining(&self) -> Option<usize> {
-        self.allocation_limit
-            .get()
-            .map(|allocation_limit| {
-                let allocated_bytes = self.allocated_bytes();
-                if allocated_bytes > allocation_limit {
-                    None
-                } else {
-                    Some(allocation_limit.abs_diff(allocated_bytes))
-                }
-            })
-            .unwrap_or(None)
+        self.allocation_limit.get().and_then(|allocation_limit| {
+            let allocated_bytes = self.allocated_bytes();
+            if allocated_bytes > allocation_limit {
+                None
+            } else {
+                Some(allocation_limit.abs_diff(allocated_bytes))
+            }
+        })
     }
 
     /// Whether a request to allocate a new chunk with a given size for a given
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -452,6 +452,13 @@ const FIRST_ALLOCATION_GOAL: usize = 1 << 9;
 // take the alignment into account.
 const DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER: usize = FIRST_ALLOCATION_GOAL - OVERHEAD;
 
+#[derive(Debug)]
+struct NewChunkMemoryDetails {
+    new_size_without_footer: usize,
+    align: usize,
+    size: usize,
+}
+
 /// Wrapper around `Layout::from_size_align` that adds debug assertions.
 #[inline]
 unsafe fn layout_from_size_align(size: usize, align: usize) -> Layout {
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -483,16 +522,105 @@ impl Bump {
...
+    }
+
+    /// How much headroom an arena has before it hits its allocation
+    /// limit.
+    #[inline(always)]
+    fn allocation_limit_remaining(&self) -> Option<usize> {
+        self.allocation_limit
+            .get()
+            .map(|allocation_limit| {
+                let allocated_bytes = self.allocated_bytes();
```
fitzgen, line 595: `allocated_bytes` has to walk over the chunks linked list to sum up all the chunk sizes, and I don't think we really want to introduce that kind of walk to new-chunk allocation. It is an `O(n)` walk, that we will do `n` times for each new chunk allocation, so it will lead to a classic `O(n^2)` behavior.

Instead, I think the `ChunkFooter` needs to start recording the total byte size of allocated chunks thus far, that is:

* the canonical empty chunk will have `allocated_bytes = 0`,
* chunk A, whose `prev = canonical empty chunk`, will have `allocated_bytes = 0 + size(A)`
* chunk B, whose `prev = A`, will have `allocated_bytes = A.allocated_bytes + size(B)`,
* etc...

Then we can have `allocated_bytes` return the current footer's `allocated_bytes` instead of walking the chunk list, and then we will be able to safely use `self.allocated_bytes()` here without introducing `O(n^2)` behavior.

The one catch is that we will need to reset our chunk's `allocated_bytes` in the `Bump::reset` method to keep it in sync with how many bytes are allocated with the `Bump`.

src/lib.rs
```diff
@@ -483,16 +522,105 @@ impl Bump {
...
+    ///
+    /// assert!(bump.try_alloc(5).is_ok());
+    /// ```
+    pub fn remove_allocation_limit(&self) {
+        self.allocation_limit.set(None)
+    }
+
+    /// How much headroom an arena has before it hits its allocation
+    /// limit.
+    #[inline(always)]
```
fitzgen, line 590: I don't think this doesn't need to be marked `inline` at all, let alone `inline(always)`. It is a method that is private to this compilation unit so LLVM will have no trouble looking at all of its callers and inlining as appropriate.

src/lib.rs
```diff
@@ -483,16 +522,105 @@ impl Bump {
...
+                    None
+                } else {
+                    Some(allocation_limit.abs_diff(allocated_bytes))
+                }
+            })
+            .unwrap_or(None)
+    }
+
+    /// Whether a new allocation fits within the headroom before the allocation limit.
+    #[inline(always)]
```
fitzgen, line 606: Same thing here.

fitzgen (commented):
Thanks! A few comments below.

src/alloc.rs
```diff
@@ -99,7 +99,7 @@ fn size_align<T>() -> (usize, usize) {
 }
 
 /// The `AllocErr` error indicates an allocation failure
-/// that may be due to resource exhaustion or to
```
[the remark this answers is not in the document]
Eliasin: Noted and reverted

src/lib.rs
```diff
@@ -483,16 +522,105 @@ impl Bump {
...
+    ///
+    /// assert!(bump.try_alloc(5).is_err());
+    ///
+    /// bump.remove_allocation_limit();
+    ///
+    /// assert!(bump.try_alloc(5).is_ok());
+    /// ```
+    pub fn remove_allocation_limit(&self) {
+        self.allocation_limit.set(None)
+    }
```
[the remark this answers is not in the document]
Eliasin: Done in latest commit

src/lib.rs
Eliasin: Removed in latest commit

src/lib.rs
Eliasin: Removed in latest commit

src/lib.rs
Eliasin: I implemented this in the latest commit, I think how I used `new_size_without_footer` to add to the `allocated_bytes` should preserve the old behavior but I'm not sure.

src/lib.rs
```diff
@@ -252,10 +252,47 @@ impl<E: Display> Display for AllocOrInitError<E> {
...
+///
+/// ### `Bump` Allocation Limits
+///
+/// `Bump` supports setting a limit on the maximum bytes of memory that can
+/// be allocated for use. This limit can be set and removed with
+/// [`set_allocation_limit`][Bump::set_allocation_limit] and
+/// [`remove_allocation_limit`][Bump::remove_allocation_limit].
+/// Changing the limit for a `Bump` while it has live allocations does
+/// nothing until a new allocation is attempted, at which point it may fail
+/// due to the new limit.
```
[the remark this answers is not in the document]
Eliasin: Suggestion committed

src/lib.rs
```diff
@@ -252,10 +252,47 @@ impl<E: Display> Display for AllocOrInitError<E> {
 /// [`Result`]: https://doc.rust-lang.org/std/result/enum.Result.html
 /// [`Ok`]: https://doc.rust-lang.org/std/result/enum.Result.html#variant.Ok
 /// [`Err`]: https://doc.rust-lang.org/std/result/enum.Result.html#variant.Err
+///
+/// ### `Bump` Allocation Limits
+///
+/// `Bump` supports setting a limit on the maximum bytes of memory that can
+/// be allocated for use. This limit can be set and removed with
```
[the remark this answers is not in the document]
Eliasin: Suggestion committed

Eliasin:
Hey @fitzgen, just checking in to see if you've had time to review the latest changes?

src/lib.rs
```diff
@@ -252,10 +252,46 @@ impl<E: Display> Display for AllocOrInitError<E> {
...
+/// ### `Bump` Allocation Limits
+///
+/// `bumpalo` supports setting a limit on the maximum bytes of memory that can
+/// be allocated for use in a particular `Bump` arena. This limit can be set and removed with
+/// [`set_allocation_limit`][Bump::set_allocation_limit].
+/// The allocation limit is only enforced when allocating new backing chunks for
+/// a `Bump`. Updating the allocation limit will not affect existing allocations
+/// or any future allocations within the `Bump`'s current chunk.
+///
+/// ## Example
```
fitzgen, line 265: This should be a sub-section of the allocation limits section, not a new super section.

```suggestion
/// #### Example
```

src/lib.rs
```diff
@@ -1316,6 +1444,13 @@ impl Bump {
     fn alloc_layout_slow(&self, layout: Layout) -> Option<NonNull<u8>> {
         unsafe {
             let size = layout.size();
+            let allocation_limit_remaining = self.allocation_limit_remaining();
+
+            // This is also checked within `new_chunk` but the check here
+            // acts as an early exit if the request could never be satisfied given the limit
+            if !Bump::new_allocation_fits_under_limit(allocation_limit_remaining, size) {
+                return None;
+            }
```
fitzgen, line 1453: We should only check this once, so either

* let's remove the check here, since `Bump::new_chunk` will check it, or
* let's remove the check in `Bump::new_chunk`, and only check it here.

I'm partial to the second option, since I think we won't need the extra parameter for `Bump::new_chunk` anymore. Although the check will have to move down into the `sizes` iterator below, to account for the size of the chunk, rather than just the size of the object (which is typically much smaller than the size of the chunk).

tests/allocation_limit.rs
```diff
@@ -0,0 +1,106 @@
...
+
+#[test]
+fn new_bump_allocated_bytes_is_zero() {
+    let bump = Bump::new();
+
+    assert_eq!(bump.allocated_bytes(), 0);
+}
+
+quickcheck! {
+    fn limit_is_never_exceeded(xs: usize) -> bool {
```
fitzgen, line 89: For clarity, can we rename this parameter?

```suggestion
    fn limit_is_never_exceeded(limit: usize) -> bool {
```

fitzgen (commented):
Thanks! Somehow this slipped through my email.

I have a few small comments below, but this generally looks good, and after those things are addressed we should be able to merge this.

Thanks for being patient!

Eliasin:
All the comments should be implemented with these commits 😄

Eliasin:
All the comments should be resolved with these commits 😄

Eliasin:
On an unrelated note, between the time I opened the PR and now I started project selection for my 4th year project in school and am looking at doing work exploring the applications of WASM modules for IoT app development and I noticed that you're a maintainer of `wasmtime`.

I'll hopefully be working with `wasmtime` and `wit-bindgen` for this so I might talk to you later on the bytecode alliance zulip 😄.

Eliasin:
> All the comments should be resolved with these commits 😄

I just have one concern regarding how allocation limits interact with the value for the minimum chunk size allocation. If a limit is set that is below `DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER`, all allocations will fail because no `sizes` will be considered that are below it.

Should `set_allocation_limit` become fallible to warn about this case? Or should a note be left in the documentation for allocation limits?

Useful documentation notes might have to effectively make the value of `DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER` public, but because of the details of this implementation, any reductions to `DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER` could be a breaking change anyway.

fitzgen:
> I just have one concern regarding how allocation limits interact with the value for the minimum chunk size allocation. If a limit is set that is below `DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER`, all allocations will fail because no `sizes` will be considered that are below it.

Ah this is a good point.

What if we chain on the limit as a size option to `new_sizes` in this scenario inside `alloc_layout_slow`:

```rust
let sizes = ...;

let sizes = sizes
    .chain(match self.allocation_limit() {
        Some(limit) if limit < DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER
            && self.allocated_bytes() == 0 => Some(limit),
        _ => None,
    });

let new_footer = ...;
```

And then the existing `filter_map` and size checking in `Bump::new` will make this Just Work.

Eliasin:
> > I just have one concern regarding how allocation limits interact with the value for the minimum chunk size allocation. If a limit is set that is below `DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER`, all allocations will fail because no `sizes` will be considered that are below it.
> 
> Ah this is a good point.
> 
> What if we chain on the limit as a size option to `new_sizes` in this scenario inside `alloc_layout_slow`:
> 
> ```rust
> let sizes = ...;
> 
> let sizes = sizes
>     .chain(match self.allocation_limit() {
>         Some(limit) if limit < DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER
>             && self.allocated_bytes() == 0 => Some(limit),
>         _ => None,
>     });
> 
> let new_footer = ...;
> ```
> 
> And then the existing `filter_map` and size checking in `Bump::new` will make this Just Work.

This mostly works but my testing finds it interacts badly with the "next power of two" rounding for small allocations and added allocation overhead, breaking the quickcheck for allocation limits never being exceeded.

To make this work I think that either the size check needs to be moved back into `new_chunk` or the logic for deciding the final layout of the allocated chunk needs to be factored out so it can be used as the size check in `alloc_layout_slow`.

Eliasin:
@fitzgen Since I'm not sure if plain comments trigger notifications

Eliasin:
I took a stab at fixing this problem by factoring out the logic for determining the final size of chunk allocations so that the code in `alloc_layout_slow` could determine the exact size of a new allocated chunk. This fixes breaking the limit because of the difference in allocation request size and the final size.

I also allowed the chunk size halving process to go below the min chunk size when dealing with small limits as just requesting a chunk with the limit value usually rounds up past the limit and ends up being rejected.

Eliasin:
I'm realizing now that this change causes some memory details about the new chunks to be calculated twice in a lot of cases so I'll make a change to avoid that now.

Eliasin:
> I'm realizing now that this change causes some memory details about the new chunks to be calculated twice in a lot of cases so I'll make a change to avoid that now.

I've made the change to cache the new chunk memory details so that they are not calculated twice but this necessitated a change in `new_chunk`, marking it as `unsafe` because of it no longer can verify (without recalculating and defeating the point) the safety of the new chunk memory details passed to it.

I'm not sure if that's an acceptable change and I'm not a fan of the name I have `NewChunkMemoryDetails` but I wanted to get a solution out and see what your opinion is.

src/lib.rs
```diff
@@ -1329,18 +1484,39 @@ impl Bump {
...
+                        true
+                    }
+                    _ => false,
+                };
+
+                if base_size >= min_new_chunk_size || bypass_min_chunk_size_for_small_limits {
                     let size = base_size;
                     base_size = base_size / 2;
-                    Some(size)
+                    Some(Bump::new_chunk_memory_details(Some(size), layout)).flatten()
```
Eliasin, line 1503: I'm not sure if `rustc` is smart enough to transform this into
```rust
if base_size >= min_new_chunk_size {
    let bypass_min_chunk_size_for_small_limits = ...;
    
    if bypass_min_chunk_size_for_small_limits {
       let size = base_size;
       base_size = base_size / 2;
       Some(Bump::new_chunk_memory_details(Some(size), layout)).flatten()
    } else {
       None
    }
```
but I decided to hold off on any premature optimizations

src/lib.rs
```diff
@@ -1329,18 +1484,39 @@ impl Bump {
...
+                        true
+                    }
+                    _ => false,
+                };
+
+                if base_size >= min_new_chunk_size || bypass_min_chunk_size_for_small_limits {
                     let size = base_size;
                     base_size = base_size / 2;
-                    Some(size)
+                    Bump::new_chunk_memory_details(Some(size), layout)
```
Eliasin, line 1503: I'm not sure if rustc is smart enough to transform this into
```rust
if base_size >= min_new_chunk_size {
    // ...
} else {
  let bypass_min_chunk_size_for_small_limits = ...;
  if bypass_min_chunk_size_for_small_limits {
     let size = base_size;
     base_size = base_size / 2;
     Bump::new_chunk_memory_details(Some(size), layout)
  } else {
    None
  }
}
```
but I decided to hold off on any premature optimizations since I think this is much less readable

tests/allocation_limit.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+#[test]
+fn small_allocation_limit() {
+    let bump = Bump::new();
+
+    // 64 is chosen because any limit below 64 is
+    // unreasonably tight when trying to optimize the
+    // allocation and will always fail due to our attempts
+    // to allocate chunks in 'nice' sizes
```
fitzgen, line 92: This comment is no longer correct.

fitzgen (commented):
Thanks! This looks good modulo a couple small things below.

src/lib.rs
```diff
@@ -471,18 +523,141 @@ impl Bump {
...
+            .get()
+            .map(|allocation_limit| {
+                let allocated_bytes = self.allocated_bytes();
+                if allocated_bytes > allocation_limit {
+                    None
+                } else {
+                    Some(allocation_limit.abs_diff(allocated_bytes))
+                }
+            })
+            .unwrap_or(None)
```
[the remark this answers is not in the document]
Eliasin: I removed this and replaced the `map` above with `and_then`.

fitzgen (approved):
Thanks! And thanks for your patience getting this one 100% over the finish line!

Eliasin:
@fitzgen Should be good now, I ran the tests on `1.54.0`

fitzgen:
Looks like CI is still failing:

https://github.com/fitzgen/bumpalo/runs/7286188380?check_suite_focus=true#step:5:342

```
failures:
---- oom_instead_of_bump_pointer_overflow stdout ----
Layout::from_size_align errored: invalid parameters to Layout::from_size_align
note: test did not panic as expected
error: test failed, to rerun pass '--test tests'
failures:
    oom_instead_of_bump_pointer_overflow
```

Eliasin:
> > Looks like CI is still failing:
> > https://github.com/fitzgen/bumpalo/runs/7286188380?check_suite_focus=true#step:5:342
> > ```
> > failures:
> > ---- oom_instead_of_bump_pointer_overflow stdout ----
> > Layout::from_size_align errored: invalid parameters to Layout::from_size_align
> > note: test did not panic as expected
> > error: test failed, to rerun pass '--test tests'
> > failures:
> >     oom_instead_of_bump_pointer_overflow
> > ```
> 
> @fitzgen I think this might be a nightly breakage as I ran this test on upstream main with nightly and experienced a failure

The test passes on this PR and upstream main on the nightly of July 10th 2022 but not July 11th 2022

Eliasin:
Toying with the latest nightly it seems that `Layout::size_from_align` panics with a lot of values that `stable` won't with. Namely values of size near `usize::MAX`. I'm not sure if this constitutes a nightly regression or where to even report something like this though

Eliasin:
I asked on the rust zulip and it seems like there is a new requirement for `Layout::from_size_align` that the size value must fit in an `isize` from this [PR](https://github.com/rust-lang/rust/pull/95295) which the [nightly docs](https://doc.rust-lang.org/nightly/std/alloc/struct.Layout.html#method.from_size_align) now show

Eliasin:
I made an issue for the breakage with a tentative suggested change

Eliasin:
Hey @fitzgen , just pinging in case this slipped by your email

fitzgen:
I don't have time to look into the old tests at the moment, if you can get them passing and send a PR then I can review that.

fitzgen:
I think this should pass CI now if rebased

Eliasin:
> I think this should pass CI now if rebased

Seems like it does!

fitzgen:
Thanks! And thanks for your patience!

Eliasin:
I had a great time working on this PR, thanks for your work reviewing it and maintaining this crate! 😄

PR #162, opened 2022-06-23
Add `boxed::Box::into_inner`

konsumlamm:
Resolves #161.

Closes #163.

PR #163, opened 2022-06-23
`bumpalo::boxed::Box::into_inner`

### LINKED ISSUE #161: Impossible to move the stored value out of the `bumpalo::Boxed::Box`.
Basically, `std::boxed::Box` allows us to move the stored value out of it just by dereferencing it:

```rust
fn foo(b: Box<AtomicI32>) -> AtomicI32 {
  *b
}
```

But equivalent code does not work for bumpalo's box:

```rust
fn foo(b: bumpalo::boxed::Box<'_, AtomicI32>) -> AtomicI32 {
  *b
}
```

Fails with
```
cannot move out of dereference of `bumpalo::boxed::Box<'_, AtomicI32>`
move occurs because value has type `AtomicI32`, which does not implement the `Copy` trait`
```

There's currently unstable associated function `std::boxed::Box::into_inner`, that allows us to do the same thing. It follows naming and some other conventions, and can be implemented easily. I propose to implement that.

```rust
fn foo(b: bumpalo::boxed::Box<'_, AtomicI32>) -> AtomicI32 {
  bumpalo::boxed::Box::into_inner(b)
}
```

ZetaNumbers:
Switches wrapped type from `&mut T` to `&mut mem::ManuallyDrop<T>`, which may improve correctness of the code, see details:
https://users.rust-lang.org/t/correctness-of-untouched-mutable-borrow-of-a-value-in-post-drop-state/77047?u=zetanumbers

Fixes #161

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -144,7 +144,7 @@ use {
 ///
 /// See the [module-level documentation][crate::boxed] for more details.
 #[repr(transparent)]
-pub struct Box<'a, T: ?Sized>(&'a mut T);
+pub struct Box<'a, T: ?Sized>(&'a mut mem::ManuallyDrop<T>);
 
 impl<'a, T> Box<'a, T> {
     /// Allocates memory on the heap and then places `x` into it.
@@ -162,14 +162,35 @@ impl<'a, T> Box<'a, T> {
     /// ```
     #[inline(always)]
     pub fn new_in(x: T, a: &'a Bump) -> Box<'a, T> {
-        Box(a.alloc(x))
+        Box(a.alloc(mem::ManuallyDrop::new(x)))
     }
 
     /// Constructs a new `Pin<Box<T>>`. If `T` does not implement `Unpin`, then
     /// `x` will be pinned in memory and unable to be moved.
     #[inline(always)]
     pub fn pin_in(x: T, a: &'a Bump) -> Pin<Box<'a, T>> {
-        Box(a.alloc(x)).into()
+        Box(a.alloc(mem::ManuallyDrop::new(x))).into()
+    }
+
+    /// Consumes the `Box`, returning the wrapped value.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use bumpalo::{Bump, boxed::Box};
+    ///
+    /// let b = Bump::new();
+    ///
+    /// let c = Box::new_in(5, &b);
+    ///
+    /// assert_eq!(Box::into_inner(c), 5);
+    /// ```
+    #[inline]
+    pub fn into_inner(boxed: Self) -> T {
+        // SAFETY: take the wrapped value and forget the box to disable drop
+        let inner = unsafe { mem::ManuallyDrop::take(boxed.0) };
+        mem::forget(boxed);
+        inner
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/boxed.rs
```diff
@@ -320,24 +341,28 @@ impl<'a, T: ?Sized> Box<'a, T> {
...
 }
 
 impl<'a> Default for Box<'a, str> {
     fn default() -> Box<'a, str> {
         // Empty slice is valid string.
         // It should be OK to `drop_in_place` empty str.
-        unsafe { Box::from_raw(Box::into_raw(Box::<[u8]>::default()) as *mut str) }
+        unsafe {
+            Box(mem::transmute::<&'a mut str, &'a mut _>(
+                core::str::from_utf8_unchecked_mut(&mut []),
```
ZetaNumbers, line 363: Actually i think i may have fixed this transmute `[u8]` to `str` part, as i am not sure you could reference anything besides the compiler source that would allow that.

konsumlamm:
Now we have two possible implementations xD (see #163). I don't think using `&mut ManuallyDrop<T>` makes a difference here, as dropping a `&mut` doesn't do anything anyway. I don't see anyone in the linked that confirms that it is safer either.

ZetaNumbers:
> I don't see anyone in the linked that confirms that it is safer either.

It is mentioned on the top, that fields of a struct must stay valid for entirety of the `Drop::drop` call.

src/boxed.rs
```diff
@@ -162,14 +162,35 @@ impl<'a, T> Box<'a, T> {
...
+    /// Consumes the `Box`, returning the wrapped value.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use bumpalo::{Bump, boxed::Box};
+    ///
+    /// let b = Bump::new();
+    ///
+    /// let c = Box::new_in(5, &b);
```
fitzgen, line 184: Can you use more descriptive variable names for these? Maybe `bump` and `five`?

src/boxed.rs
fitzgen: I don't follow what you're talking about here. Why was the previous code not acceptable?

fitzgen (commented):
FWIW, I don't see the harm in a `ManuallyDrop` here, even if it isn't 100% clear it is necessary (and nothing in those discussions or linked threads is telling me this is super necessary either, so maybe you could link the specific replies you are alluding to @ZetaNumbers?)

src/boxed.rs
konsumlamm, line 184: I think it would be better to use a non-`Copy` type that implements `Drop` here (e.g. `String`), so that we also test that we don't run the destructor twice (Miri checks that, at least if we enable the `boxed` feature in the CI).

src/boxed.rs
ZetaNumbers: I believe `str`'s underlying layout is not defined.

fitzgen: Feel free to add a safety comment, I think the original code can remain otherwise unchanged tho.

src/boxed.rs
fitzgen: Ah, I bet they wouldn't mind a PR giving their example better names as well then!

konsumlamm:
@ZetaNumbers do you still plan to work on this?

ZetaNumbers:
> @ZetaNumbers do you still plan to work on this?

Yes. Sorry, got a bit busy with my own project and then got somewhat sick. Let me try doing this today actually.

ZetaNumbers:
I remember that adding `ManuallyDrop` is not sufficient for variance reasons. It's best to use `NonNull` instead to allow covariance. So i believe this PR is no better than #162.

konsumlamm:
Do you want me to revive #162 or is that just an observation?

ZetaNumbers:
Sure, please borrow anything you want from here.

PR #168, opened 2022-06-27
Fix `cargo miri test --all-features`

### LINKED ISSUE #164: `cargo miri test --features collections` fails
Currently, the CI only runs Miri without any features enabled. When I tried to enable all features, I got an "Undefined Behavior" error, see https://github.com/fitzgen/bumpalo/runs/7022998455?check_suite_focus=true.

### LINKED ISSUE #166: `Vec::set_len` example is UB
While i hope it's not, it seems intentional. If it is... why?
https://github.com/fitzgen/bumpalo/blob/fa52577688284d73b44cc8f49671ef02ebdea14b/src/collections/vec.rs#L1028-L1042

ZetaNumbers:
- [x] `-Zmiri-ignore-leaks` baseline
- [x] `-Zmiri-tag-raw-pointers`
- [x] `-Zmiri-strict-provenance` (`cargo miri --version` f76ebd6 2022-07-02)
- [ ] eliminate `-Zmiri-ignore-leaks`

Also adds `test_skip_miri_quickchecks` default feature.

Fixes #166, fixes #164


ZetaNumbers:
So i ran `cargo miri test --test quickchecks`, and it have spuriously failed on `alloc_strs` and `can_allocate_big_values` tests, only indicating that one of its spawned `cargo miri` process exited with error code 255. Could this be OOM bc of a leak?

fitzgen:
I doubt there is a leak here, since I think valgrind (which we run tests under in CI) would have complained about that.

Cargo.toml
```diff
@@ -35,10 +35,12 @@ criterion = "0.3.0"
 rand = "0.7"
 
 [features]
-default = []
+default = ["test_skip_miri_quickchecks"]
 collections = []
 boxed = []
 allocator_api = []
+# quickchecks may take a long time to run in MIRI
+test_skip_miri_quickchecks = []
```
fitzgen, line 43: Can we not use cargo features for this? Doesn't MIRI set some sort of `cfg` that we can use instead?

tests/quickchecks.rs
```diff
@@ -1,9 +1,7 @@
-// Miri is very slow. Even with QUICKCHECK_MAX_TESTS=1 the tests in this module run for more than
-// an hour.
```
ZetaNumbers, line 2: Actually environment variables are invisible to miri, unless you disable its isolation or pass environment variable at compile time. Properly enabling `QUICKCHECK_TESTS=1` actually increases `cargo miri test --all-features` time from ~1:40 to only ~4:00, so i enabled one quickcheck run on miri by default.

tests/collect_in.rs
```diff
@@ -1,28 +1,24 @@
 #![cfg(feature = "collections")]
+
+mod quickcheck;
```
fitzgen, line 3: Instead of making every test program recompile the `quickcheck` module, let's reorganize the tests like

```
bumpalo/
  - tests/
    - all/
      - main.rs
      - quickcheck.rs
      - etc...
```

and `main.rs` has everything else as submodules (maybe `cfg`ed on or off depending what features are enabled and all that).

This way, we only compile the `quickcheck` module once, and don't need to have many `mod quickcheck` declarations.

tests/quickcheck.rs
```diff
@@ -0,0 +1,34 @@
...
+                // Use QUICKCHECK_TESTS from compiletime to surpass miri isolation
+                let tests = option_env!("QUICKCHECK_TESTS").and_then(|s| s.parse().ok());
+                #[cfg(miri)]
+                let tests = tests.or(Some(1));
+
+                let mut qc = ::quickcheck::QuickCheck::new();
+
+                if let Some(tests) = tests {
+                    qc = qc.tests(tests)
+                }
```
fitzgen, line 28: Rather than defaulting to 1 if `QUICKCHECK_TESTS` is not supplied, can we default to whatever `quickcheck`'s default is, and set `QUICKCHECK_TESTS=1` in the CI env for MIRI?

fitzgen (commented):
Thanks! A couple comments below.

tests/quickcheck.rs
ZetaNumbers: Well, yes. That's what this code does.

tests/collect_in.rs
ZetaNumbers: I don't see how you would avoid these "recompilations" anyway, if we keep tests separated inside of multiple files, since you need to `mod something` for each of these files. I don't think joining file would be a good idea too, i just tried to do what [the rust book says](https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests). I could take the `common/mod.rs` advice tho...

fitzgen: Each `.rs` file under `tests/` is compiled as its own crate (including the module without any tests!). By having a module that is used by all of them, it is recompiled N times, once for each crate. With the layout I proposed, there is only one crate, and the module is therefore just compiled once. It might not be the hugest deal at the moment, but it sets us up for the growth of these tests in the future.

ZetaNumbers: > Each `.rs` file under `tests/` is compiled as its own crate (including the module without any tests!).

Yes, and it doesn't recurse into a directory. By moving `tests/quickcheck.rs` into `tests/quickcheck/mod.rs` in [0f538b1](https://github.com/fitzgen/bumpalo/pull/168/commits/0f538b11af7f25719585e11c005246f32dd0b388) i was able to delete this empty `quickcheck` test.

Keep in mind this tests structure is recommended by the rust book and, i believe, should be considered standard. Until this code works fine If you see an improvement to be made by changing tests structure, then i would ask you to handle it via a separate PR, since it would be outside of the scope of this PR. And one more thing, i would recommend not to do that restructure unless you detect a significant improvement in build time, disk usage, or something else.

ZetaNumbers: Also recompilations like these are common in C++ world with header-only libraries there.

fitzgen: The C++ world is really a non sequitur here. Doesn't have any relevance.

The `tests/all/main.rs` structure is a common idiom:

* https://github.com/bytecodealliance/wasmtime/blob/main/tests/all/main.rs
* https://github.com/rustwasm/wasm-bindgen/blob/main/tests/wasm/main.rs
* https://github.com/rust-fuzz/cargo-fuzz/blob/master/tests/tests/main.rs
* it is even used in `cargo` itself: https://github.com/rust-lang/cargo/blob/master/tests/testsuite/main.rs

but like. man. I'm tired. we've been going back and forth on this for too long. I don't have the energy to keep this up. it's my repo. just structure it the way I prefer. thanks.

ZetaNumbers: Implemented in 161b65009096b18d20077cbb3e9fa09464e5e611

fitzgen:
Thank you for your patience and for sticking with this PR!

ZetaNumbers:
No problem. Sorry for the long delay.

PR #176, opened 2022-08-03
Fix oom_instead_of_bump_pointer_overflow breakage

### LINKED ISSUE #174: `oom_instead_of_bump_pointer_overflow` test breakage on `nightly-2022-07-11`
The `oom_instead_of_bump_pointer_overflow` test fails on `nightly-2022-07-11` because of an [upstream change](https://github.com/rust-lang/rust/pull/95295) to `rustc` to restrict the size passed to `Layout::from_size_align` to be within `isize::MAX` instead of `usize::MAX`. Since this test uses `usize::MAX - p + 1` as a parameter to ensure overflow, this test now fails because of the call to `Layout::from_size_align` failing.

Eliasin:
Nightly changes tp the `Layout` API have caused this test to break as
layouts are now restricted to a max size of `isize::MAX`. This is fixed
by using `isize::MAX` instead of `usize::MAX` in the test.

This test has been a bit controversial in the past but some of the problems
with it have been dealt with up to the present, like not allocating enough to
guarantee an OOM error.

Fixes #174

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -71,7 +71,7 @@ fn oom_instead_of_bump_pointer_overflow() {
     let p = x as *mut u8 as usize;
 
     // A size guaranteed to overflow the bump pointer.
-    let size = usize::MAX - p + 1;
+    let size = (isize::MAX as usize) - p + 1;
     let align = 1;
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Thanks!

Issue #178, opened 2022-08-12
Use after free in `bumpalo::Vec::into_iter`


uselessgoddess:
Miri humiliate the following code:
```rust
use bumpalo::{vec, Bump};

fn foo() -> impl Iterator<Item = usize> {
    let arena = Bump::new();
    vec![in &arena; 1, 2, 3].into_iter()
}

fn main() {
    for x in foo() {
        println!("{x}");
    }
}
```
```
error: Undefined Behavior: pointer to alloc2451 was dereferenced after this allocation got freed
2207 |                 self.ptr = self.ptr.offset(1);
     |                            ^^^^^^^^^^^^^^^^^^ pointer to alloc2451 was dereferenced after this allocation got freed
     |
     = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
     = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
     = note: backtrace:
note: inside `main` at src\main.rs:16:14
    --> src\main.rs:16:14
     |
16   |     for x in foo() {
     |              ^^^^^
```
Why is that? Burrow into bumpalo `into_iter` implementations.
```rust
// Hm... unsafe. Interesting
unsafe {
    let begin = self.as_mut_ptr();
    // assume(!begin.is_null());
    let end = if mem::size_of::<T>() == 0 {
        arith_offset(begin as *const i8, self.len() as isize) as *const T
    } else {
        // oh no, we forgot the borrowing checker
        begin.add(self.len()) as *const T
    };
    mem::forget(self);
    IntoIter {
        phantom: PhantomData,
        ptr: begin,
        end,
    }
}
```
```rust
fn foo() -> impl Iterator<Item = usize> {
    let arena = Bump::new();
    vec![in &arena; 1, 2, 3].into_iter()
    // <-- bump drop here and free our memory 
    // but ptrs in `IntoIter` still point to this memory
}
```

konsumlamm:
See also #164 and #168.

konsumlamm:
The problem is that this code shouldn't compile in the first place. I think the solution is to add a lifetime parameter to `IntoIter` (and all the other consuming iterators).

uselessgoddess:
I think it would be interesting if it were possible to return an `IntoIter` that would extend the lifetime of the `Bump`

fitzgen:
Fixed in #181

shinmao:
[...]

PR #181, opened 2022-10-14
vec::IntoIter fixes

main--:
Ran into these issues while tracking down a memory leak.

### DIFF
--- a/src/collections/vec.rs
+++ b/src/collections/vec.rs
@@ -1941,7 +1941,7 @@ impl<'bump, T: 'bump> ops::DerefMut for Vec<'bump, T> {
 
 impl<'bump, T: 'bump> IntoIterator for Vec<'bump, T> {
     type Item = T;
-    type IntoIter = IntoIter<T>;
+    type IntoIter = IntoIter<'bump, T>;
 
     /// Creates a consuming iterator, that is, one that moves each value out of
     /// the vector (from start to end). The vector cannot be used after calling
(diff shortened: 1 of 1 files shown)


konsumlamm:
This fixes #178.

src/collections/vec.rs
```diff
@@ -2341,9 +2341,16 @@ impl<'bump, T: 'bump> DoubleEndedIterator for IntoIter<T> {
...
 
-impl<'bump, T: 'bump> FusedIterator for IntoIter<T> {}
+impl<'bump, T: 'bump> FusedIterator for IntoIter<'bump, T> {}
+
+impl<'bump, T> Drop for IntoIter<'bump, T> {
+    fn drop(&mut self) {
+        // drop all remaining elements
+        self.for_each(drop);
+    }
+}
```
fitzgen, line 2353: Can this use a drop guard the same way that `alloc` does?

https://cs.github.com/rust-lang/rust/blob/9b0a099dfc9a97ecd10adb319396c731c4b2d169/library/alloc/src/vec/into_iter.rs?q=drop+for+intoiter#L355-L377

fitzgen (commented):
Thanks! One comment to be addressed before this can merge.

src/collections/vec.rs
main--: If my understanding of the implementation you linked is correct, then no. It seems to me that the idea there is that if an element destructor were to panic then at least the Vec’s allocation would still be freed.

This should already work the same way in bumpalo, given that the backing allocation belongs to the Bump struct, which would still be dropped normally during unwinding, if IntoIter’s drop were to panic.

fitzgen (approved):
Thanks again!

PR #188, opened 2022-11-28
Use ManuallyDrop with bumpalo's Box instead of mem::forget

saethlin:
Miri now reports UB with some uses of bumpalo's Box where it did not before: https://github.com/rust-lang/miri/issues/2704

The previous behavior of Miri was a false negative. rustc applies noalias to newtype wrappers around &mut, so Miri has to retag &mut when passed by value to a function even if it is in a wrapper struct, such as bumpalo's Box. mem::forget is a common aliasing footgun, because for a unique-owning wrapper like Box, leaking an RAII handle re-asserts the uniqueness of the handle as it is sent to the void. Ouch.

ManuallyDrop solves this problem.

Closes https://github.com/fitzgen/bumpalo/issues/187

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -130,7 +130,7 @@ use {
             future::Future,
             hash::{Hash, Hasher},
             iter::FusedIterator,
-            mem,
+            mem::ManuallyDrop,
             ops::{Deref, DerefMut},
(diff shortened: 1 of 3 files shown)


fitzgen (approved):
Thanks!

PR #189, opened 2023-01-06
Add Vec::bump method

mattfbacon:
This PR adds a way to get the `Bump` allocator backing a `Vec`, for example to make more allocations if the `Vec` contains other bump-allocated data structures.

One common example of this is a bump-allocated recursive data structure. Consider:

```rust
use bumpalo::Bump;
use bumpalo::collections::Vec;

struct Tree<'bump> {
	value: i32,
	children: Vec<'bump, Self>,
}

impl Tree<'bump> {
	fn new(value: i32, bump: &'bump Bump) -> Self {
		Self {
			value,
			children: Vec::new_in(bump),
		}
	}

	fn add_child(&mut self, value: i32) {
		self.children.push(Self::new(value, self.children.bump()));
	}
}
```

Without this PR, `Recursive` would be forced to carry an extra shared reference to the `Bump`.

### DIFF
--- a/src/collections/vec.rs
+++ b/src/collections/vec.rs
@@ -675,6 +675,26 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
         }
     }
 
+    /// Returns a shared reference to the allocator backing this `Vec`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use bumpalo::{Bump, collections::Vec};
+    ///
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Thanks!

Would you be interested in adding `Box::bump` and `String::bump` as well? Note the the box version should be a static method to match the `std` `Box::allocator` method.

PR #197, opened 2023-03-13
Add `#[inline]` annotations to all public functions

Kmeakin:
Functions which are not marked `#[inline]` cannot be inlined when called from third-party crates without LTO enabled. This PR adds `#[inline]` annotations to public functions which were missing them, and adds a clippy warning to ensure future public functions are always marked `#[inline]`

### DIFF
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -109,6 +109,7 @@ pub struct AllocErr;
 // (we need this for downstream impl of trait Error)
 // #[unstable(feature = "allocator_api", issue = "32838")]
 impl fmt::Display for AllocErr {
+    #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.write_str("memory allocation failed")
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,7 @@
 #![doc = include_str!("../README.md")]
 #![deny(missing_debug_implementations)]
 #![deny(missing_docs)]
+#![deny(clippy::missing_inline_in_public_items)]
 #![no_std]
 #![cfg_attr(
     feature = "allocator_api",
@@ -44,11 +45,13 @@ pub enum AllocOrInitError<E> {
     Init(E),
 }
 impl<E> From<AllocErr> for AllocOrInitError<E> {
+    #[inline]
     fn from(e: AllocErr) -> Self {
         Self::Alloc(e)
     }
 }
 impl<E: Display> Display for AllocOrInitError<E> {
+    #[inline]
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
             AllocOrInitError::Alloc(err) => err.fmt(f),
@@ -353,6 +356,7 @@ static EMPTY_CHUNK: EmptyChunkFooter = EmptyChunkFooter(ChunkFooter {
 });
 
 impl EmptyChunkFooter {
+    #[inline]
     fn get(&'static self) -> NonNull<ChunkFooter> {
         unsafe { NonNull::new_unchecked(&self.0 as *const ChunkFooter as *mut ChunkFooter) }
     }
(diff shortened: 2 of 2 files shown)


src/alloc.rs
```diff
@@ -109,6 +109,7 @@ pub struct AllocErr;
 // (we need this for downstream impl of trait Error)
 // #[unstable(feature = "allocator_api", issue = "32838")]
 impl fmt::Display for AllocErr {
+    #[inline]
```
fitzgen, line 112: This is not a performance sensitive function, and shouldn't need to be inlined.

src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![doc = include_str!("../README.md")]
 #![deny(missing_debug_implementations)]
 #![deny(missing_docs)]
+#![deny(clippy::missing_inline_in_public_items)]
```
fitzgen, line 4: I don't use clippy, nor do I believe this particular rule is useful to apply 100% of the time, so I don't think it makes sense to add this.

src/lib.rs
```diff
@@ -44,11 +45,13 @@ pub enum AllocOrInitError<E> {
     Init(E),
 }
 impl<E> From<AllocErr> for AllocOrInitError<E> {
+    #[inline]
     fn from(e: AllocErr) -> Self {
         Self::Alloc(e)
     }
 }
 impl<E: Display> Display for AllocOrInitError<E> {
+    #[inline]
```
fitzgen, line 54: This isn't performance sensitive either.

src/lib.rs
```diff
@@ -353,6 +356,7 @@ static EMPTY_CHUNK: EmptyChunkFooter = EmptyChunkFooter(ChunkFooter {
 });
 
 impl EmptyChunkFooter {
+    #[inline]
```
fitzgen, line 359: This isn't even a public function. Were you seeing that it was not being inlined when profiling or something?

src/lib.rs
```diff
@@ -361,6 +365,7 @@ impl EmptyChunkFooter {
 impl ChunkFooter {
     // Returns the start and length of the currently allocated region of this
     // chunk.
+    #[inline]
```
fitzgen, line 368: Same question here.

src/lib.rs
```diff
@@ -371,18 +376,21 @@ impl ChunkFooter {
     }
 
     /// Is this chunk the last empty chunk?
+    #[inline]
```
fitzgen, line 379: And here.

src/lib.rs
```diff
@@ -1856,8 +1883,7 @@ impl Iterator for ChunkRawIter<'_> {
 
 impl iter::FusedIterator for ChunkRawIter<'_> {}
 
-#[inline(never)]
-#[cold]
+#[inline]
```
fitzgen, line 1886: This should never be inlined, the whole point is that it is a very cold path, that is out of lined. Function call overhead is negligible compared to the cost of panicking, and inlining this code will slow down the vast majority common case when there is no OOM.

fitzgen (changes requested):
Thanks for taking the time to make a pull request.

Happy to make `pub` things `inline` where it makes sense, but some of these don't make sense (are not `pub` -- in which case I'd like to know  that the annotations are profiling driven -- or are for things that are clearly not performance sensitive like display/debug implementations).

The vibe I'm getting is that you added a clippy lint and then blindly followed it, which is an approach that avoids considering context and understanding of why some things are the way they are (e.g. applying this rule to functions that are intentionally marked never inline) and ultimately wastes both of our time.

If you have actual profiles showing that some hot methods are not being inlined, I will accept a PR adding those specific inline annotations. I don't want to spend time reviewing PRs that make absolutely everything inline for no reason.

src/lib.rs
Kmeakin: `panic!()` expands to a call to `core::panicking::panic`, which is itself marked `#[inline(never)]`

Kmeakin:
The motivation was initially to try and remove function calls from the emitted assembly to calls like `Bump::new()`, and `Bump::alloc()`, but yes I then discovered the clippy lint and applied it everywhere it suggested. I will close this issue and reopen a new one if I can find any profile differences.

PR #201, opened 2023-04-05
Use allocator-api through allocator-api2 crate

zakarumych:
This change allows to use mirror of the allocator-api on stable channel.
Types imported from `allocator-api2` are reexports from `core` and `alloc` crates when `"nightly"` feature is enabled.
Otherwise `allocator-api2` defines types to mimic unstable `core` and `alloc` types and functions, stable stuff is re-exported.

The goal is to add support for `allocator-api2` into as many crates as possible.
The `bumpalo` crate is first on the list.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,7 +2,7 @@
 #![deny(missing_debug_implementations)]
 #![deny(missing_docs)]
 #![no_std]
-#![cfg_attr(feature = "nightly", feature(allocator_api))]
+#![cfg_attr(feature = "allocator_api", feature(allocator_api))]
 
 #[doc(hidden)]
 pub extern crate alloc as core_alloc;
@@ -24,10 +24,10 @@ use core::slice;
 use core::str;
 use core_alloc::alloc::{alloc, dealloc, Layout};
 
-#[cfg(feature = "nightly")]
+#[cfg(feature = "allocator_api")]
 use core_alloc::alloc::{AllocError, Allocator};
 
-#[cfg(all(feature = "allocator_api", not(feature = "nightly")))]
+#[cfg(all(feature = "allocator-api2", not(feature = "allocator_api")))]
 use allocator_api2::alloc::{AllocError, Allocator};
 
 pub use alloc::AllocErr;
@@ -1887,7 +1887,7 @@ unsafe impl<'a> alloc::Alloc for &'a Bump {
     }
 }
 
-#[cfg(feature = "allocator_api")]
+#[cfg(any(feature = "allocator_api", feature = "allocator-api2"))]
 unsafe impl<'a> Allocator for &'a Bump {
     fn allocate(&self, layout: Layout) -> Result<NonNull<[u8]>, AllocError> {
         self.try_alloc_layout(layout)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,10 +2,7 @@
 #![deny(missing_debug_implementations)]
 #![deny(missing_docs)]
 #![no_std]
-#![cfg_attr(
-    feature = "allocator_api",
-    feature(allocator_api, nonnull_slice_from_raw_parts)
-)]
+#![cfg_attr(feature = "allocator_api", feature(allocator_api))]
 
 #[doc(hidden)]
 pub extern crate alloc as core_alloc;
@@ -25,9 +22,8 @@ use core::mem;
 use core::ptr::{self, NonNull};
 use core::slice;
 use core::str;
-use core_alloc::alloc::{alloc, dealloc, Layout};
-#[cfg(feature = "allocator_api")]
-use core_alloc::alloc::{AllocError, Allocator};
+
+use allocator_api2::alloc::{alloc, dealloc, AllocError, Allocator, Layout};
 
 pub use alloc::AllocErr;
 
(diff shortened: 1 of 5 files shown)


zakarumych:
Apparently this would require bumping MSRV to 1.63

src/lib.rs
```diff
@@ -1920,7 +1919,9 @@ unsafe impl<'a> Allocator for &'a Bump {
         new_layout: Layout,
     ) -> Result<NonNull<[u8]>, AllocError> {
         Bump::grow(self, ptr, old_layout, new_layout)
-            .map(|p| NonNull::slice_from_raw_parts(p, new_layout.size()))
+            .map(|p| unsafe {
+                NonNull::new_unchecked(ptr::slice_from_raw_parts_mut(p.as_ptr(), new_layout.size()))
+            })
```
fitzgen, line 1924: Why these changes? Seems unnecessary?

fitzgen (commented):
Thanks for the PR!

Can we make the dependency optional? Right now, `bumpalo` doesn't have any mandatory dependencies, and I'd like to keep it that way.

Additionally, once the dependency is optional, we should have a blurb in the README just before/after the existing blurb for the `allocator_api` feature about what enabling it does.

src/lib.rs
zakarumych: To remove all unstable rust features. Except that I had to put one back, so this one may stay as well I guess

zakarumych:
How do you suggest enabling "nightly" in `allocator-api2`?

zakarumych:
I found a problem with the approach in `allocator-api2` - replacing stable mirror API with unstable API re-export makes all dependent crates manually enable unstable `feature(allocator_api)`.
If some other crate would enable `"allocator-api2/nightly"` feature and not enable `"bumpalo/nightly"` then `bumpalo` won't compile.

So I changed it.
`allocator-api2` is now always stable mirror and crates choose from using `allocator-api2` and `alloc` crate, picking `alloc` when they enable nightly Rust feature `allocator_api`.

zakarumych:
I tried to forward "nightly" feature back from `allocator-api2` into `bumpalo`, but it requires other unstable features.

src/lib.rs
```diff
@@ -2,7 +2,7 @@
 #![deny(missing_debug_implementations)]
 #![deny(missing_docs)]
 #![no_std]
-#![cfg_attr(feature = "allocator_api", feature(allocator_api))]
+#![cfg_attr(feature = "nightly", feature(allocator_api))]
```
fitzgen: Ideally, the API and cargo features for enabling the nightly allocator API support in `bumpalo` wouldn't change just because we can alternatively use the `allocator_api2` crate, so the "nightly" feature should keep the "allocator_api" name.

src/lib.rs
```diff
@@ -1886,6 +1891,7 @@ unsafe impl<'a> alloc::Alloc for &'a Bump {
     }
 }
 
+#[cfg(feature = "allocator_api")]
```
fitzgen: This should be

```rust
#[cfg(any(feature = "allocator_api", feature = "allocator_api2"))]
```

fitzgen (commented):
Aside: it seems like it would make sense for `allocator_api2` to have a cargo feature that, when enabled, made the crate re-export the `std` types/traits rather than polyfilling them. I don't think `bumpalo` woulds use that feature, but it would help avoid ecosystem splitting and annoying build errors where some crates are using the actual `Allocator` and others are using `allocator_api2::Allocator`.

zakarumych:
It was made like this originally.

However this doesn't work since user would have to enable unstable #![feature(allocator_api)] if any crate enables "nightly" in allocator-api2.

I found no way to enable #![feature(allocator_api)] conditionally based on allocator-api2 features.
Tried using macro, but inner macro attributes are unstable.
It's also impossible to get custom cfg flag from dependency.

I tried different route and make blanket impls of allocator_api2::Allocator for core::alloc::Allocator types.
This way allocators that implement core::alloc::Allocator would be compatible.
Allocators that use allocator-api2 would implement core::alloc::Allocator with "nightly" feature as well.

Unfortunately I got stuck on conflicting impls for references. Tried using specialization feature to resolve the conflict, but with no luck.

So currently PR is blocked on resolving this issue. I'll return to it later when I have free time again.

zakarumych:
Ok. I think it should be consistent and usable now.
`allocator-api2` re-exports `core`, `alloc` and `std` with "nightly" feature.
Other crates that use it should have means to enable `#![feature(allocator_api)]` to avoid compilation errors on nightly channel.
`bumpalo` does. Even if direct dependent crates won't enable it since it doesn't care about nightly, indirect dependent crates would be able to enable it if they encounter compilation error when some other crate in the tree enable "allocator-api2/nightly".

fitzgen (approved):
Looks great, thanks for your patience as we worked out the exact kinks of how the cargo features and `cfg`s would work!

fitzgen:
It does look like this needs a little rebase to deal with `CHANGELOG.md` conflicts, and in that case can you also remove the now-obsolete bits in the changelog about the "nightly" feature? I was going to do that myself, but since it needs a rebase anyways...

zakarumych:
It's ready for merge now

fitzgen (approved):
Thanks!!

zakarumych:
@fitzgen can you provide ETA for next release with this?

fitzgen:
Published!

PR #202, opened 2023-04-05
Remove `unsafe` from `EmptyChunkFooter::get()`

Johan-Mi:
Have I completely misunderstood this or can we really get rid of an `unsafe` block this easily?


fitzgen (approved):
Good eye, thanks!

PR #203, opened 2023-04-08
Add `Bump::alloc_str_concat`

SabrinaJewson:
I thought this would be a useful function to have. The motivation behind the `&impl AsRef<str>` design is that it greatly simplifies the implementation in comparison to `impl AsRef<str>` (see the previous commit) — also, as the iterator is iterated twice users should be producing references anyway.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1269,6 +1269,10 @@ impl Bump {
         let mut bytes = iter.flat_map(|s| s.as_ref().bytes());
         let buffer = self.alloc_slice_fill_with(total_len, |_| bytes.next().unwrap());
 
+        // This assert is necessary to guard against misbehaving `Iterator` implementations leading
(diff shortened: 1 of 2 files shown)


fitzgen (commented):
Thanks for making a PR.

However, I'm not convinced this method belongs in this crate? It doesn't seem like a super common use case, and since it can be implemented entirely with the existing API that this crate exposes it doesn't seem like it *has* to be built in?

SabrinaJewson:
That’s true — I was motivated to make this PR by the fact that it does require some `unsafe` code that is non-trivial to get right. The thing to me is that the majority of cases where people would benefit from this method they’ll end up using `.alloc_str(&format!` instead which is much slower — on the other hand, if this method is supported they’ll just have faster code.

fitzgen:
Thanks for opening this PR, but after some consideration I don't think it belongs here.

PR #209, opened 2023-07-07
Fix chunk_capacity to return the remaining capacity, not the total

msullivan:
`chunk_capacity` is documented as returning "the remaining capacity in the current chunk", but actually returns the *total* capacity for the current chunk.

(Arguably, the actual behavior is itself totally sensible, and this is a documentation bug, though in that case I would propose adding a new function that returns the remaining capacity. I would like it for debugging/measurement purposes.)


fitzgen (approved):
Thanks!

Issue #63, opened 2020-01-21
Collections compatibility with Serde


rw:
Would it be feasible to have the bumpalo Vec type work with serde? That way we can handle deserializing variable length sequences of references to borrowed data. That would be a big improvement over using the global allocator.

Here are three example types. The first is not Deserialize with serde, because it requires borrowing a slice of references to strings, which is a variable-length data structure. The second is Deserialize with serde, but it causes expensive heap allocations. The third could be possible using bumpalo.

```rust
#[derive(Serialize, Deserialize)]
struct NotSerde<'a> {
    items: &'a [&'a str],
}

struct YesSerdeButRequiresHeapAllocation<'a> {
    #[serde(borrow)]
    items: Vec<&'a str>,
}

struct YesSerdeWithBumpalo<'a> {
    #[serde(borrow)]
    items: bumpalo::collections::Vec<'a, &'a str>,
}
```

fitzgen:
A new `"serde"` feature seems reasonable to me. When enabled, it would turn on serialize and deserialize impls for everything in `bumpalo::collections`. I think we would need to implement them by hand (or at least some of them) so I think it makes sense to avoid enabling the derive feature of serde in bumpalo.

zakarumych:
Note that you'd have to implement only `serde::de::DeserializeSeed` to provide reference to `Bump` instance.

chloekek:
It should be noted that the `Deserialize` impl for `Vec` cannot always use `with_capacity_in`, because `SeqAccess::size_hint` may return `None`. This can cause excessive memory allocation when deserializing large sequences, depending on the `Deserializer` impl.

jaredh159:
@fitzgen hey, i took a whack at implementing this today, but i think the deserialize side is inherently problematic due to the recursive nature of deriving deserialize.  you can use `DeserializeSeed`, but i think you're sort of stuck at one level, if you have some struct that has a `Vec` in it, or some deeper nesting, i don't think there is a way to materialize a `Bump` out of thin air while recursing and handle the lifetime. the serde crate seems to acknowledge this limitation, see https://github.com/serde-rs/serde/issues/1325 and dtolnay's code and comments here: https://github.com/serde-rs/serde/issues/1325.

[...]

PR #210, opened 2023-07-13
add serde serialization support

keithamus:
This starts work on #63 by adding support for `Serialize` on `Box` and `Vec`, with some basic tests & docs.

I couldn't figure out the right way to implement `DeserializeSeed` for `Vec`. I'm a newcommer to rust so I think I need some pointers on how to handle that, as I'm struggling with lifetimes:

```rs

#[cfg(feature = "serde")]
impl<'a, T> DeserializeSeed<'a> for Vec<'a, T>
where
    T: Deserialize<'a>,
{
    type Value = ();

    fn deserialize<D>(mut self, deserializer: D) -> Result<Self::Value, D::Error>
    where
        D: Deserializer<'a>,
    {
        struct VecVisitor<'a, T: 'a>(&'a mut Vec<'a, T>);

        impl<'a, T> Visitor<'a> for VecVisitor<'a, T>
        where
            T: Deserialize<'a>,
        {
            type Value = ();

            fn expecting(
                &self,
                formatter: &mut core_alloc::fmt::Formatter,
            ) -> core_alloc::fmt::Result {
                Ok(())
            }

            fn visit_seq<V>(self, mut visitor: V) -> Result<(), V::Error>
            where
                V: SeqAccess<'a>,
            {
                while let Some(elem) = visitor.next_element()? {
                    self.0.push(elem);
                }
                Ok(())
            }
        }

        deserializer.deserialize_seq(VecVisitor::<'a>(self.as_mut()))
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,37 @@ in its space itself.
 }
 ```
 
+#### Serde
+
+Adding the `serde` feature flag will enable transparent serialization of Vecs and 
+boxed values.
+
+```toml
+[dependencies]
+bumpalo = { version = "3.9", features = ["collections", "boxed", "serde"] }
+```
+
+```rust,ignore
+use bumpalo::{Bump, boxed::Box, collections::Vec};
+
+// Create a new bump arena.
+let bump = Bump::new();
+
+// Create a `Box`
+let box = Box::new_in("hello", &bump);
+
+// Serialize with serde_json
+assert_eq!(serde_json::to_string(&box).unwrap(), "\"hello\"");
+
+// Create a `Vec`
+let vec = Vec::new_in( &bump);
+vec.push(1);
+vec.push(2);
+
+// Serialize with serde_json
+assert_eq!(serde_json::to_string(&vec).unwrap(), "[1, 2]");
+```
+
 ### `#![no_std]` Support
 
 Bumpalo is a `no_std` crate by default. It depends only on the `alloc` and `core` crates.
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -681,3 +681,19 @@ impl<'a, T, const N: usize> TryFrom<Box<'a, [T]>> for Box<'a, [T; N]> {
         }
     }
 }
+
+#[cfg(feature = "serde")]
+mod serialize {
+    use super::*;
+
+    use serde::{Serialize, Serializer};
+
+    impl<'a, T> Serialize for Box<'a, T>
+    where
+        T: Serialize,
+    {
+        fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+            T::serialize(self, serializer)
+        }
+    }
+}
--- a/src/collections/vec.rs
+++ b/src/collections/vec.rs
@@ -2760,3 +2760,27 @@ impl<'bump> io::Write for Vec<'bump, u8> {
         Ok(())
     }
 }
+
+#[cfg(feature = "serde")]
+mod serialize {
+    use super::*;
+
+    use serde::{ser::SerializeSeq, Serialize, Serializer};
+
+    impl<'a, T> Serialize for Vec<'a, T>
+    where
+        T: Serialize,
+    {
+        fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: Serializer,
+        {
+            let mut seq = serializer.serialize_seq(Some(self.len))?;
+            for e in self.iter() {
+                seq.serialize_element(e)?;
+            }
+            seq.end()
+        }
+    }
+
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,10 +36,15 @@ harness = false
 # implement its `Allocator` trait.
 allocator-api2 = { version = "0.2.8", default-features = false, optional = true }
 
+# This dependency is here to allow integration with Serde, if the `serde` feature is enabled
+serde = { version = "1.0.171", optional = true }
+
 [dev-dependencies]
 quickcheck = "1.0.3"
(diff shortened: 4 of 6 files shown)


README.md
```diff
@@ -102,7 +102,32 @@ arenas.
...
+let vec = Vec::new_in( &bump);
+vec.push(1);
+vec.push(2);
+
+// Serialize with serde_json
+assert_eq!(serde_json::to_string(&vec).unwrap(), "[1, 2]");
+```
+
+Eventually
+[all `std` collection types will be parameterized by an
```
fitzgen, line 130: Can you leave the "Eventually ..." paragraph with the first part of the `### Collections` section, and have the `#### Serde` section follow after that paragraph?

fitzgen: Actually, can we just add a sentence/paragraph about serde here, rather than a whole subsection, and then reference/link to a dedicated/deduplicated serde section down below. That way we won't need to mention serde in both the collections and box sections.

src/boxed.rs
```diff
@@ -681,3 +681,16 @@ impl<'a, T, const N: usize> TryFrom<Box<'a, [T]>> for Box<'a, [T; N]> {
...
+
+#[cfg(feature = "serde")]
+impl<'a, T> Serialize for Box<'a, T>
+where
+    T: Serialize,
+{
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        (**self).serialize(serializer)
+    }
+}
```
fitzgen, line 696: Can we put this in a `mod` for tidiness?

```rust
#[cfg(feature = "serde")]
mod serialize {
    use super::*;
    use serde::{Serialize, Serializer};

    // the serialize impl...
}
```

src/boxed.rs
```diff
@@ -681,3 +681,16 @@ impl<'a, T, const N: usize> TryFrom<Box<'a, [T]>> for Box<'a, [T; N]> {
...
+#[cfg(feature = "serde")]
+use serde::{Serialize, Serializer};
+
+#[cfg(feature = "serde")]
+impl<'a, T> Serialize for Box<'a, T>
+where
+    T: Serialize,
+{
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        (**self).serialize(serializer)
```
fitzgen, line 694: Can this use unified function call syntax so it is easier for readers to verify that this isn't infinite looping at a glance?

```suggestion
        T::serialize(self, serializer)
```

src/collections/vec.rs
```diff
@@ -2760,3 +2760,23 @@ impl<'bump> io::Write for Vec<'bump, u8> {
...
+    where
+        S: Serializer,
+    {
+        let mut seq = serializer.serialize_seq(Some(self.len))?;
+        for e in self.iter() {
+            seq.serialize_element(e)?;
+        }
+        seq.end()
+    }
+}
```
fitzgen, line 2782: Ditto about a module for tidiness.

tests/all/boxed.rs
```diff
@@ -12,3 +12,13 @@ fn into_raw_aliasing() {
...
 }
+
+#[cfg(feature = "serde")]
+#[test]
+fn test_box_serializes() {
+    let bump = Bump::new();
+    let boxed = Box::new_in(1, &bump);
+    assert_eq!(serde_json::to_string(&boxed).unwrap(), "1");
+    let boxedStr = Box::new_in("a", &bump);
+    assert_eq!(serde_json::to_string(&boxedStr).unwrap(), "\"a\"");
```
fitzgen, line 23: Can we also test a few other things:

* tuple
* struct w/ a couple fields
* enum w/ a couple variants
* vec w/ a couple elements
* hash map
* hash set

It should be fine to add a dev dependency on `serde_derive` if it makes this easier, fwiw.

Also, no need to explicitly write out the expected JSON; should be able to call `serde_json::to_string(&unboxed_value)` and then `serde_json::to_string(&boxed_value)` and compare the results for equality, I believe. Or compare `std::boxed::Box<T>` with `bumpalo::boxed::Box<'_, T>`. Might be worth writing a little helper that does this dance, rather than writing it out in full for all the cases above.

tests/all/vec.rs
```diff
@@ -167,3 +167,16 @@ fn test_vec_write() {
...
+fn test_vec_serializes() {
+    let bump = Bump::new();
+    let mut vec = Vec::new_in(&bump);
+    vec.push("hello");
+    vec.push("world");
+    assert_eq!(
+        serde_json::to_string(&vec).unwrap(),
+        "[\"hello\",\"world\"]"
+    );
+}
```
fitzgen, line 182: Similarly, can we sanity test more cases than just `Vec<&'static str>`? And can we compare against `std::vec::Vec<T>`'s serialization rather than hard coding the JSON?

Also worth deserializing the JSON into a `std::vec::Vec<T>` and asserting that the values round tripped properly. (No use in adding serialization if nothing can deserialize it.) Same for the box case.

fitzgen (commented):
Thanks! And sorry for the delay giving feedback on this. A handful of comments below.

README.md
keithamus: Sorry this was a rebase error.

tests/all/boxed.rs
keithamus: I've added a struct with an enum and option. Hopefully that's sufficient, or would you like me to add more such as the hash map/set?

tests/all/vec.rs
keithamus: Okay I've added more test cases and tested them against serialize/deserialize with `std::vec::Vec`

fitzgen (approved):
Thanks!

PR #213, opened 2023-08-16
deps: Update criterion to 0.5

waywardmonkeys:


fitzgen (approved):
Thanks! Does this version of criterion still produce HTML reports by default, or do we need to enable a feature now?

waywardmonkeys:
I will get back to you about that (in 18-36 hours as I'm about to disappear for 16 hours or so). There's also the issue of the failed MSRV build. Arguably, the MSRV build shouldn't compile everything, but only the core library?

waywardmonkeys:
@fitzgen Okay, I added `html_reports` to the features for it.

That leaves MSRV. `anstyle` requires 1.64, you want 1.63 (but the `rust-version` in the `Cargo.toml` is still set to 1.60). Would you want to bump to 1.64 or 1.66? Or not compile tests for MSRV?

fitzgen:
The tests should run for the MSRV, but benchmarks don't need to. I think cargo compiles all dev deps for tests though.

I think it is fine to bump the MSRV. Would you mind updating the `rust-version` in `Cargo.toml`, CI, and the `README`'s documentation as well?

Thanks @waywardmonkeys!

waywardmonkeys:
Now `anstyle` wants 1. 70 or later ... I'm not going to chase this as it is somewhat ridiculous.

fitzgen:
Closing old pull requests.

Issue #215, opened 2023-09-05
Implement `std::io::Write` for `Vec<'bump, u8>`


marmeladema:
First of all, thank you for bumpalo, it's really awesome :)

I encountered a roadblock for a specific usecase recently when I wanted to use `serde_json::to_writer` on a `Vec<'bump, u8>` when I realized `std::io::Write` is not actually implemented whereas it is for `Vec<u8>` in the standard library: https://doc.rust-lang.org/std/io/trait.Write.html#impl-Write-for-Vec%3Cu8,+A%3E

There does not seem to be any technical limitation in bumpalo that would prevent from implementing it except:
```
Bumpalo is a no_std crate. It depends only on the alloc and core crates.
```
since `std::io::Write` is not available in `core`.

@fitzgen would you be inclined to accept a contribution that implement `std::io::Write`? Probably gated by a new `std` feature disabled by default?

Thank you for taking the time to read my request.

fitzgen:
Yes, introducing a `std` feature that is disabled by default makes sense. Make sure to document the feature in the readme as well. Thanks!

PR #216, opened 2023-09-14
Implement `std::io::Write` for `Vec<'bump, u8>`

marmeladema:
Hi @fitzgen 

Here is my attempt to implement `std::io::Write` for `Vec<'bump, u8>`
I believe there are other common traits like this we could implement for some collections but I have left that for another time (but I 'd be happy to contribute those).

I have a few questions:
1. The current `std` feature does not do anything on its own, it's currently only useful if you also have the `collections` feature enabled too. I think it's fine but I wondered if having instead a `collections-std` feature depending on `collections` would be better? Although the `std` feature is pretty standard across the community.
2. Are there enough tests and are they at the right place? The implementation seems trivial and has been copied from std (except for the vectored part that is currently unstable).

Fixes https://github.com/fitzgen/bumpalo/issues/215

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ default = []
 collections = []
 boxed = []
 allocator_api = []
+std = []
 
 # [profile.bench]
 # debug = true
--- a/README.md
+++ b/README.md
@@ -155,7 +155,14 @@ in its space itself.
 
 ### `#![no_std]` Support
 
-Bumpalo is a `no_std` crate. It depends only on the `alloc` and `core` crates.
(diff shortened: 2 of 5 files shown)


fitzgen (approved):
Thanks!

Yeah, it is fine that `std` without `collections` doesn't do anything. That's the community norm name and it gives us a place to add any non-collection related `std` things in the future if it is ever needed.

marmeladema:
If that's not too much to ask, that would be helpful yeah.

fitzgen:
Just published 3.14.0!

marmeladema:
Thanks a lot! That's really helpful :-)

PR #217, opened 2023-09-24
fixed `_` instead of `-` in name `allocator-api2`

mgerhold:
The names of both the referenced crate and the cargo feature are `allocator-api2` and *not* `allocator_api2`.

This change was introduced [here](https://github.com/fitzgen/bumpalo/commit/fd3dd5767658ad634581d0bc8267c3ed00ec9ec5#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R212-R213).

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,8 +216,8 @@ v.push(2);
 
 ### Using the `Allocator` API on Stable Rust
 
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Thansk!

PR #218, opened 2023-09-25
clippy: Allow large_enum_variant in tests.

waywardmonkeys:
2 instances of this enum already had this, but it was missing from these two.


fitzgen (commented):
Thanks for the PR! However, I don't personally use clippy because I find that the suggestions are not always useful / are sometimes too opinionated, and I would rather not litter the code base with annotations like these just to satisfy it.

fitzgen:
I'll remove the existing annotations.

PR #223, opened 2023-11-17
Annotate `Allocator` methods with `#[inline]`

Kmeakin:
These methods were not annotated with `#[inline]`, meaning they will not be inlined into client crates unless they enable LTO

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1875,6 +1875,7 @@ unsafe impl<'a> alloc::Alloc for &'a Bump {
 
 #[cfg(any(feature = "allocator_api", feature = "allocator-api2"))]
 unsafe impl<'a> Allocator for &'a Bump {
+    #[inline]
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Thanks!

fitzgen:
Mind bumping the MSRV to 1.65.0 while you're here? It looks like `regex` now requires that so the tests aren't passing on 1.63.0 anymore.

waywardmonkeys:
I ran into the MSRV thing with #224 as well. I'll submit a separate PR that fixes that.

fitzgen:
#225 merged so this can be rebased on that and then we can merge it.

Kmeakin:
Ready to merge. Sorry it wasn't done sooner, I'd forgotten this patch was pending.

fitzgen:
Thanks!

PR #227, opened 2024-02-04
Replace modulo ops with bitwise-and equivalent

cynecx:
Doesn't really affect any of the in-tree benchmarks due to the statically known allocation sizes and heavy inlining. LLVM is smart enough to optimize the expensive modulo operation away. However in more dynamic cases this produces less and more optimal code.


fitzgen (commented):
We already have a `round_down_to` helper function that can be called here, and which debug asserts its correctness predicates. Can you update the PR to make use of that helper? Thanks!

cynecx:
> We already have a round_down_to helper function that can be called here...

I don't think we can use the existing helpers because they don't preserve pointer provenance due to operating on `usize`. Do you want me to add this equivalent helper instead?

```rust
#[inline]
pub(crate) fn round_ptr_down_to(ptr: *const u8, divisor: usize) -> *const u8 {
    debug_assert!(divisor > 0);
    debug_assert!(divisor.is_power_of_two());
    ptr.wrapping_sub(ptr as usize & (divisor - 1))
}
```

Or keep this as is?

fitzgen:
Let's add the equivalent helper, thanks!

cynecx:
🙏 I've updated the branch :)

fitzgen (approved):
Thanks!

src/lib.rs
```diff
@@ -420,6 +420,13 @@ pub(crate) fn round_down_to(n: usize, divisor: usize) -> usize {
     n & !(divisor - 1)
 }
 
+#[inline]
```
fitzgen, line 423:
```suggestion
/// Same as `round_down_to` but preserves pointer provenance.
#[inline]
```

Issue #226, opened 2024-01-18
`Box::pin_in` violates pin's drop guarantee


zetanumbers:
This is incorrect and violates pin's [drop guarantee](https://doc.rust-lang.org/std/pin/index.html#drop-guarantee) which in turn breaks existing correct code assuming this rule. For this reason [`std::box::Box::pin_in`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.pin_in) has `A: 'static` bound over the allocator generic type argument.

zetanumbers:
https://github.com/rust-lang/rust/pull/79327#issue-748392410

> Allocators has to retain their validity until the instance and all of its clones are dropped. When pinning a value, it must live forever, thus, the allocator requires a `'static` lifetime for pinning a value. [Example from reddit](https://www.reddit.com/r/rust/comments/jymzdw/the_story_continues_vec_now_supports_custom/gd7qak2?utm_source=share&utm_medium=web2x&context=3):
> 
> ```rust
> let alloc = MyAlloc(/* ... */);
> let pinned = Box::pin_in(42, alloc);
> mem::forget(pinned); // Now `value` must live forever
> // Otherwise `Pin`'s invariants are violated, storage invalidated
> // before Drop was called.
> // borrow of `memory` can end here, there is no value keeping it.
> drop(alloc); // Oh, value doesn't live forever.
> ```

Issue #186, opened 2022-11-08
Pinning `bumpalo::boxed::Box` is unsound


Lej77:
A pinned `bumpalo::boxed::Box` might not uphold `Pin`'s [Drop guarantee](https://doc.rust-lang.org/std/pin/index.html#drop-guarantee). Specifically if a pinned `bumpalo::boxed::Box` is forgotten then the memory backing it might still be freed. (This differs from how `std::box::Box` is implemented.)

To soundly give out pinned references from a `Bump` one could keep a counter of the number of pinned allocations that still need to be freed and if that counter isn't zero when `Bump` is dropped then it could leak all of its memory. This counter would require `bumpalo::boxed::Box` to store a reference to `Bump` however which would double the `Box`'s size (two pointers).

khoover:
This combined with #204 makes me think we want two box types; a `ThinBox` that's still the single word, but doesn't hold on to the `Bump` pointer, and a `FatBox` that does and can be `Pin`ned or `Clone`d.

maxdexh:
@fitzgen since you mentioned elsewhere that you're making a major release with breaking changes soon, maybe you can remove `pin_in` and `From<Box> for Pin` as well?

PR #228, opened 2024-02-04
Pinned Box 

cynecx:
Fixes https://github.com/fitzgen/bumpalo/issues/226, https://github.com/fitzgen/bumpalo/issues/186.

This adds new "pinned" Box type (`pin::Box<T>`), that guarantees running `T`'s drop impl whenever the box is leaked and the `Bump` allocator is reset or dropped.

The impl utilizes a "intrusive circular doubly linked list" that links every "pinned" allocation that needs a cleanup. Each list entry detaches when the Box itself is dropped.
That way, when the `Bump` allocator is cleared, we can run the `Drop` impls by traversing the list.

The pinned Box is still one pointer-sized big. We can calculate the each allocation's header pointer by doing some pointer arithmetic. Note that this only passes miri with tree-borrows enabled (not compatible with stacked borrows and reports UB) due to the kind of container-of style pointer arithmetic being used. 

# TODO

- Some more cleanups
- Right now the docs are basically copy and pasted and needs to be rewritten
- More testing

### DIFF
--- a/src/pin.rs
+++ b/src/pin.rs
@@ -167,6 +167,7 @@ impl<'a, T> Box<'a, T> {
 
     /// Constructs a new `Pin<Box<T>>`. If `T` does not implement `Unpin`, then
     /// `x` will be pinned in memory and unable to be moved.
+    #[inline]
     pub fn pin_in(x: T, a: &'a Bump) -> Pin<Box<'a, T>> {
         Box::new_in(x, a).into()
     }
@@ -278,11 +279,58 @@ impl<'a, T: ?Sized> Box<'a, T> {
     /// }
     /// ```
     #[inline]
-    pub fn into_raw(b: Box<'a, T>) -> *mut T {
+    pub fn into_raw(b: Box<'a, T>) -> *mut T
+    where
+        T: Unpin,
+    {
+        unsafe { Box::into_raw_unchecked(b) }
+    }
+
+    /// Consumes the `Box`, returning a wrapped raw pointer.
+    ///
+    /// The pointer will be properly aligned and non-null.
+    ///
+    /// After calling this function, the caller is responsible for the
+    /// value previously managed by the `Box`. In particular, the
+    /// caller should properly destroy `T`. The easiest way to
+    /// do this is to convert the raw pointer back into a `Box` with the
+    /// [`Box::from_raw`] function, allowing the `Box` destructor to perform
+    /// the cleanup.
+    ///
+    /// Note: this is an associated function, which means that you have
+    /// to call it as `Box::into_raw(b)` instead of `b.into_raw()`. This
+    /// is so that there is no conflict with a method on the inner type.
+    ///
+    /// # Examples
+    ///
+    /// Converting the raw pointer back into a `Box` with [`Box::from_raw`]
+    /// for automatic cleanup:
+    /// ```
+    /// use bumpalo::{Bump, pin::Box};
+    ///
+    /// let b = Bump::new();
+    ///
+    /// let x = Box::new_in(String::from("Hello"), &b);
+    /// let ptr = unsafe { Box::into_raw_unchecked(x) };
+    /// let x = unsafe { Box::from_raw(ptr, &b) }; // Note that new `x`'s lifetime is unbound. It must be bound to the `b` immutable borrow before `b` is reset.
+    /// ```
+    /// Manual cleanup by explicitly running the destructor:
+    /// ```
+    /// use std::ptr;
+    /// use bumpalo::{Bump, pin::Box};
+    ///
+    /// let b = Bump::new();
+    ///
+    /// let mut x = Box::new_in(String::from("Hello"), &b);
+    /// let p = unsafe { Box::into_raw_unchecked(x) };
+    /// unsafe {
+    ///     ptr::drop_in_place(p);
+    /// }
+    /// ```
+    #[inline]
+    pub unsafe fn into_raw_unchecked(b: Box<'a, T>) -> *mut T {
         let b = ManuallyDrop::new(b);
-        unsafe {
-            b.release_from_drop_list();
-        }
+        b.release_from_drop_list();
         b.ptr
     }
 
@@ -396,7 +444,7 @@ impl<'a, T: ?Sized + Hasher> Hasher for Box<'a, T> {
     }
 }
 
-impl<'a, T> From<Box<'a, T>> for Pin<Box<'a, T>> {
+impl<'a, T: ?Sized> From<Box<'a, T>> for Pin<Box<'a, T>> {
     /// Converts a `Box<T>` into a `Pin<Box<T>>`.
     ///
     /// This conversion does not allocate on the heap and happens in place.
--- a/src/pin.rs
+++ b/src/pin.rs
@@ -9,7 +9,7 @@
 //! Move a value from the stack to the heap by creating a [`Box`]:
 //!
 //! ```
-//! use bumpalo::{Bump, boxed::Box};
+//! use bumpalo::{Bump, pin::Box};
 //!
 //! let b = Bump::new();
 //!
@@ -20,7 +20,7 @@
 //! Move a value from a [`Box`] back to the stack by [dereferencing]:
 //!
 //! ```
-//! use bumpalo::{Bump, boxed::Box};
+//! use bumpalo::{Bump, pin::Box};
 //!
 //! let b = Bump::new();
 //!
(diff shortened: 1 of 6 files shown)


fitzgen:
I don't have time to dig in yet, but here are some initial thoughts:

* I don't mind losing the reset-is-just-a-pointer-update property because (a) that isn't quite true already when there are multiple chunks and (b) presumably if the user doesn't allocate any pinned boxes in the arena, then it doesn't impose any overhead other than a single compare and branch on the empty intrusive list, which is fine. If the implementation doesn't reflect (b) then we should investigate ways to make it so.

* I have vague plans for adding sub-arenas/checkpoints in the next breaking release of bumpalo (that said, nothing is concretely planned yet, and it might even never happen). This would take advantage of the LIFO nature of bump allocation to reset the bump pointer to a place in the middle of a chunk where the user made a checkpoint, keeping alive earlier allocations but freeing everything allocated since that checkpoint. To be compatible with this, the drop list would need to be stored in youngest-to-oldest allocation order so that we could keep dropping pinned boxes while they were past the checkpoint and stop as soon as we encounter one that is within the checkpoint. If the list was in the opposite order, resetting to a checkpoint would need to walk over every pinned box within the checkpoint before getting to the ones to drop that are outside the checkpoint, effectively making the operation O(total pin boxes) rather than O(pin boxes to drop). Is this ordering requirement problematic? I don't think it should be, but I also haven't looked into the details yet.

* It would be very good to have extensive quickcheck tests for this stuff.

* > Note that this only passes miri with tree-borrows enabled (not compatible with stacked borrows and reports UB) due to the kind of container-of style pointer arithmetic being used.

  Is this fundamental to the intrusive list design or an implementation detail? All else being equal it would be nice to support both models, but it wouldn't be the end of the world to support only tree borrows. Would it help to hold a pointer to the header inside the box and then access the data as an offset of the header, rather than the other way around? I don't know if that would better propagate provenance or whatever.

cynecx:
> I don't mind losing the reset-is-just-a-pointer-update property

👍

> presumably if the user doesn't allocate any pinned boxes in the arena, then it doesn't impose any overhead other...

Yes, that's how it's currently implemented. (Besides, the sentinel node is also currently lazily allocated and initialized.)

> I have vague plans for adding sub-arenas/checkpoints in the next breaking release of bumpalo...

Exciting stuff!

> Is this ordering requirement problematic? I don't think it should be, but I also haven't looked into the details yet.

Yeah, I don't think the ordering requirement is going to be a problem. The current impl uses a doubly linked list, so you could just iterate forwards or backwards regardless in what order you push the drop entries.

However, the way some APIs are currently implemented might pose an issue here:

Both `pin::Box<T>::into_raw` and `pin::Box<T>::from_raw` unlink/re-add the drop entry when called.
Therefore, invalidating the ordering invariant that we'd possibly want to preserve.
Perhaps the question is rather whether these APIs should've done that in the first place.
Or perhaps we should just keep the drop entry attached, and store a `is_detached: bool` in the header and toggle that when `into_raw/from_raw` is called.

Also, I am not exactly sure about the design you have in mind but I can also imagine one could maintain multiple drop lists for each sub-arena/checkpoint,
but that kind of depends on how it's actually implemented...

> It would be very good to have extensive quickcheck tests for this stuff.

I've added some basic tests that test the API surface of the pinned Box. But we should probably add some tests around the internals (drop list) too.

> Is this fundamental to the intrusive list design or an implementation detail?

It's more of an implementation detail (See below).

> Would it help to hold a pointer to the header inside the box and then access the data as an offset of the header, rather than the other way around?

That won't work with SB because it's still doing "pointer arithmetic" ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5c1730527241cfaa8b5773598da3e530)). IIUC, pointers can only be derived by actually "reborrowing" from a pointer,
aka, we'd have to repr the pinned Box like `struct Box<T>(*const DropEntry<T>)` and derive the data pointer with something like `&(*self.0).data`.
However, that representation is somewhat wonky because unsized coercion doesn't work anymore, which is quite unfortunate because we can't have a `Box<dyn Future>` anymore.

Although, I admit I'm not an expert in the SB/Tree-Borrows pointer aliasing models, so I might be totallt wrong about this 😅.

cynecx:
> I don't mind losing the reset-is-just-a-pointer-update property

👍

> presumably if the user doesn't allocate any pinned boxes in the arena, then it doesn't impose any overhead other...

Yes, that's how it's currently implemented. (Besides, the sentinel node is also currently lazily allocated and initialized.)

> I have vague plans for adding sub-arenas/checkpoints in the next breaking release of bumpalo...

Exciting stuff!

> Is this ordering requirement problematic? I don't think it should be, but I also haven't looked into the details yet.

Yeah, I don't think the ordering requirement is going to be a problem. The current impl uses a doubly linked list, so you could just iterate forwards or backwards regardless in what order you push the drop entries.

However, the way some APIs are currently implemented might pose an issue here:

Both `pin::Box<T>::into_raw` and `pin::Box<T>::from_raw` unlink/re-add the drop entry when called.
Therefore, invalidating the ordering invariant that we'd possibly want to preserve.
Perhaps the question is rather whether these APIs should've done that in the first place.
Or perhaps we should just keep the drop entry attached, and store a `is_detached: bool` in the header and toggle that when `into_raw/from_raw` is called.

Also, I am not exactly sure about the design you have in mind but I can also imagine one could maintain multiple drop lists for each sub-arena/checkpoint,
but that kind of depends on how it's actually implemented...

> It would be very good to have extensive quickcheck tests for this stuff.

I've added some basic tests that test the API surface of the pinned Box. But we should probably add some tests around the internals (drop list) too.

> Is this fundamental to the intrusive list design or an implementation detail?

It's more of an implementation detail (See below).

> Would it help to hold a pointer to the header inside the box and then access the data as an offset of the header, rather than the other way around?

That won't work with SB because it's still doing "pointer arithmetic" ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e5e87bc79398b4d79fa1d278bb8c0c7c)). IIUC, pointers can only be derived by actually "reborrowing" from a pointer,
aka, we'd have to repr the pinned Box like `struct Box<T>(*const DropEntry<T>)` and derive the data pointer with something like `&(*self.0).data`.
However, that representation is somewhat wonky because unsized coercion doesn't work anymore, which is quite unfortunate because we can't have a `Box<dyn Future>` anymore.

Although, I admit I'm not an expert in the SB/Tree-Borrows pointer aliasing models, so I might be totallt wrong about this 😅.

src/pin.rs
```diff
@@ -0,0 +1,663 @@
...
+/// implementations.
+///
+/// See the [module-level documentation][crate::boxed] for more details.
+#[repr(transparent)]
+pub struct Box<'a, T: ?Sized> {
+    ptr: *mut T,
+    _marker: PhantomData<&'a mut T>,
+}
+
+impl<'a, T> Box<'a, T> {
```
zetanumbers, line 153: `T` must satisfy `T: 'a` (or be able to outlive `'a`) so that drop within the bump would not be called on invalid value.

```suggestion
impl<'a, T> Box<'a, T>
where
    T: 'a
{
```

zetanumbers (changes requested):

src/pin.rs
zetanumbers: Actually `'a` is just an lifetime to an immutable reference of `Bump`, which can be subtyped into smaller lifetimes via covariance, so this is not good enough. Appropriate fix for would be to add a lifetime argument `'b` to `Bump` itself to restrict it and then restrict `T: 'b`.

cynecx: The marker `PhantomData<&'a mut T>` inside `Box<'a, T>` should produce the implied bound `T: 'a`, so your suggested change should be a noop.

> Actually 'a is just an lifetime to an immutable reference of Bump, which can be subtyped into smaller lifetimes via covariance, so this is not good enough.

I don't see how this is unsound. Even if you'd have `&'short Bump`, `T` must outlive `'short` (and `'long: 'short`). The returned ref to `T` will be borrowed for the shorter lifetime and can't live longer than that.

It would be great if you could provide an example where this unsoundness is demonstrated :)

zetanumbers: You drop `T: 'short` object within the `Bump: 'long`'s drop.

cynecx: Sorry, I don't quite understand what you are trying to say. It would really help the discussion if you could provide a quick poc of the unsoundness.

Aside, since this `pinned::Box` implementation is almost an 1:1 interface copy of the existing `boxed::Box` implementation, your claim also affects the current in-tree `Box` implementation. So again, I'd be really interested in seeing this unsoundness you are talking about.

cynecx: Ah I see the issue now. Yeh, it's actually very easy to trigger with something like

```rust
#[test]
fn pinned_failure() {
    struct Foo<'a>(&'a String);

    impl<'a> Drop for Foo<'a> {
        fn drop(&mut self) {
            println!("{}", self.0);
        }
    }

    let bump = Bump::new();

    let s = "hello".to_string();
    let p = Box::new_in(Foo(&s), &bump);
    mem::forget(p);
    drop(s);

    drop(bump);
}
```

PR #232, opened 2024-02-14
Fix `Bump::shrink` illegal copies

### LINKED ISSUE #230: `allocator_api` tests failing
The CI failure discovered in #229 is also failing on `main` (reproduced locally on M1 Macbook Pro).

`cargo test --features allocator_api` on nightly is sufficient to trigger it.

The failing test is: https://github.com/fitzgen/bumpalo/blob/f8278d6fff870d34bda7c2d03173710b2c9d12ea/tests/all/allocator_api.rs#L163-L211

Have traced it to: https://github.com/fitzgen/bumpalo/blob/f8278d6fff870d34bda7c2d03173710b2c9d12ea/src/lib.rs#L1705-L1745

That test seems to produce a lot of calls to `Bump::shrink` where `old_size` and `new_size` are both 1, and `ptr` and `new_ptr` are equal (so the copy is indeed overlapping).

In these cases, `delta == 0` and `old_size / 2 == 0`, so `(delta >= old_size / 2) == true`.

Changing that test to `delta > 0 && delta >= old_size / 2` solves this specific problem (`cargo test --all-features` passes), but I'm not sure if it still would allow other illegal cases through.

I'm afraid I don't understand the logic well enough to make a PR, but I hope this may be helpful in finding a fix.

overlookmotel:
Fixes #230.

Or at least I think I've understood what the problem is here. Previously could end up with overlapping copies when `old_size` is an odd number.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1724,7 +1724,10 @@ impl Bump {
                 // Only reclaim the excess space (which requires a copy) if it
                 // is worth it: we are actually going to recover "enough" space
                 // and we can do a non-overlapping copy.
-                && delta >= old_size / 2
+                // `(old_size + 1) / 2` so division rounds up rather than down,
+                // so this test fails for e.g. `old_size` = 1, `new_size` = 1
+                // or `old_size` = 5, `new_size` = 3 (which would overlap).
+                && delta >= (old_size + 1) / 2
         {
             let footer = self.current_chunk_footer.get();
             let footer = footer.as_ref();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1724,9 +1724,31 @@ impl Bump {
                 // Only reclaim the excess space (which requires a copy) if it
                 // is worth it: we are actually going to recover "enough" space
                 // and we can do a non-overlapping copy.
-                // `(old_size + 1) / 2` so division rounds up rather than down,
-                // so this test fails for e.g. `old_size` = 1, `new_size` = 1
-                // or `old_size` = 5, `new_size` = 3 (which would overlap).
+                //
+                // We do `(old_size + 1) / 2` so division rounds up rather than
+                // down. Consider when:
+                //
+                //     old_size = 5
+                //     new_size = 3
+                //
+                // If we do not take care to round up, this will result in:
+                //
+                //     delta = 2
+                //     (old_size / 2) = (5 / 2) = 2
+                //
+                // And the the check will succeed even though we are have
+                // overlapping ranges:
+                //
+                //     |--------old-allocation-------|
+                //     |------from-------|
+                //                 |-------to--------|
+                //     +-----+-----+-----+-----+-----+
+                //     |  a  |  b  |  c  |  .  |  .  |
+                //     +-----+-----+-----+-----+-----+
+                //
+                // But we MUST NOT have overlapping ranges because we use
+                // `copy_nonoverlapping` below! Therefore, we round the
+                // division up to avoid this issue.
                 && delta >= (old_size + 1) / 2
         {
             let footer = self.current_chunk_footer.get();


src/lib.rs
```diff
@@ -1724,7 +1724,10 @@ impl Bump {
                 // Only reclaim the excess space (which requires a copy) if it
                 // is worth it: we are actually going to recover "enough" space
                 // and we can do a non-overlapping copy.
-                && delta >= old_size / 2
+                // `(old_size + 1) / 2` so division rounds up rather than down,
+                // so this test fails for e.g. `old_size` = 1, `new_size` = 1
+                // or `old_size` = 5, `new_size` = 3 (which would overlap).
```
fitzgen, line 1729: `old_size = 1; new_size=1` would result in `delta = 0`, so I don't think it is a good example to use here.

The other example is good, since it actually shows the bug. How about we expand the comment a little more like this:

```suggestion
                //
                // We do `(old_size + 1) / 2` so division rounds up rather than
                // down. Consider when:
                //
                //     old_size = 5
                //     new_size = 3
                //
                // If we do not take care to round up, this will result in:
                //
                //     delta = 2
                //     (old_size / 2) = (5 / 2) = 2
                //
                // And the the check will succeed even though we are have
                // overlapping ranges:
                //
                //     |--------old-allocation-------|
                //     |------from-------|
                //                 |-------to--------|
                //     +-----+-----+-----+-----+-----+
                //     |  a  |  b  |  c  |  .  |  .  |
                //     +-----+-----+-----+-----+-----+
                //
                // But we MUST NOT have overlapping ranges because we use
                // `copy_nonoverlapping` below! Therefore, we round the
                // division up to avoid this issue.
```

fitzgen (commented):
Thanks for finding and fixing this bug!

I've expanded on the comment as a suggestion below. I'll just commit that directly.

I think we should additionally do the following:

* Debug assert that the ranges are not overlapping before we call `copy_nonoverlapping`.

* Write a new quickcheck test for explicitly testing grow/shrink. This would take a `Vec<((usize, usize), (usize, usize))>` input where each pair is treated as a `(size, align)` and clamped and rounded to pow2 as necessary. We would allocate the first pair's layout, then realloc to the second pair's layout. This should thoroughly exercise the growing and shrinking methods.

Would you be up for sending a follow up pull request to implement these additional two things?

overlookmotel:
2 questions on the first:

Do you mean for all `copy_nonoverlapping` usage in bumpalo, or just the specific one in `shrink`?

Is `debug_assert!` required? In debug mode, it seems that `copy_nonoverlapping` does exactly that anyway (hence the test failure):
https://doc.rust-lang.org/src/core/intrinsics.rs.html#2684-2704

fitzgen:
The quickcheck tests live in `tests/all/quickchecks.rs`. Here is a simple quickcheck test: https://github.com/fitzgen/bumpalo/blob/7b3fd17f73b48ed874ec68ebc7970354c8c74be8/tests/all/quickchecks.rs#L231-L237 

It is "property-based testing" (which is pretty much the same as fuzzing, but kind of a separate community due to historical accidents) where you write a "property" which is a function that takes some input, runs your program/library on it, and asserts something about the result. The quickcheck library itself is responsible for generating a bunch of inputs and running them through the property. In bumpalo, our properties are often just "no assertions were triggered when allocating the given stuff", but there are a few others if you read through the tests.

overlookmotel:
I'm sorry, but I'm not having luck with this. I don't think I understand the logic of when can shrink or grow, while also changing alignment.

This is where I got to, but it fails on the case `((2, 0), (0, 2))`:

```rs
fn allocator_grow_or_shrink(layouts: Vec<((usize, usize), (usize, usize))>) -> bool {
    fn to_layout(size: usize, align: usize) -> Layout {
        const MIN_SIZE: usize = 1;
        const MAX_SIZE: usize = 1024;
        const MIN_ALIGN: usize = 1;
        const MAX_ALIGN: usize = 64;

        let align = usize::min(usize::max(align, MIN_ALIGN), MAX_ALIGN);
        let align = usize::next_power_of_two(align);

        let size = usize::min(usize::max(size, MIN_SIZE), MAX_SIZE);
        let size = size.next_multiple_of(align);

        Layout::from_size_align(size, align).unwrap()
    }

    let mut layout_iter = layouts.into_iter();
    while let Some(((initial_size, initial_align), (new_size, new_align))) = layout_iter.next() {
        let initial_layout = to_layout(initial_size, initial_align);
        let new_layout = to_layout(new_size, new_align);

        let b = AllocatorDebug::new(Bump::new());

        let pointer = b.allocate(initial_layout).unwrap();
        if !is_pointer_aligned_to(pointer, initial_layout.align()) {
            return false;
        }

        let new_pointer = if new_layout.size() <= initial_layout.size() {
            unsafe { b.shrink(pointer.cast(), initial_layout, new_layout) }.unwrap()
        } else {
            unsafe { b.grow(pointer.cast(), initial_layout, new_layout) }.unwrap()
        };

        if !is_pointer_aligned_to(new_pointer, new_layout.align()) {
            return false;
        }
    }

    true
}
```

Also, the test `allocator_shrink_align_change` includes:

```rs
let size = usize::min(usize::max(size, MIN_SIZE), MAX_SIZE);
let size = usize::max(size, align);
```

But don't all allocations have to have `size` a multiple of `align` (i.e. `size = size.next_multiple_of(align)`) not just `size >= align`?

I'm afraid I think I need to bail out at this point! I only really wanted to get #229 merged, and made this fix to get CI to pass. But I'm afraid I'm now a bit short of time to continue on. Sorry.

PR #233, opened 2024-02-15
Add a quickcheck for growing or shrinking existing allocations

fitzgen:
cc @overlookmotel: just an FYI in case you were interested in how I ended up doing this.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -14,7 +14,7 @@ jobs:
   build:
     strategy:
       matrix:
-        rust_channel: ["stable", "beta", "nightly", "1.65.0"]
+        rust_channel: ["stable", "beta", "nightly", "1.73.0"]
         feature_set: ["--features collections,boxed"]
         include:
           - rust_channel: "nightly"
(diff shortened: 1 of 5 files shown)


overlookmotel:
I *am* interested. Thanks for pinging me.

Right, so the actual logic around grow/shrink and alignment had to change a bit, not just the test.

fitzgen:
That or alternatively the test could have accepted `Err(_)` in the case where requested new alignment was greater than the old alignment, but I figured I'd loosen things up a little while I was here.

overlookmotel:
Nice. Yes, I guess no reason to refuse to increase alignment, if the address is already aligned.

Can I ask you one random question that has nothing to do with this issue? (just don't want to open an issue just for it)

Is the reason for filling the buffers backwards (start at the back, and move towards the front) to make alignment calculations more efficient?

fitzgen:
> Is the reason for filling the buffers backwards (start at the back, and move towards the front) to make alignment calculations more efficient?

I wrote a whole blog post about this :)

https://fitzgeraldnick.com/2019/11/01/always-bump-downwards.html

overlookmotel:
Haha oops! Well I guess that answers my question (in spades). Thank you, will digest.

Issue #235, opened 2024-02-17
Improve performance of `extend_from_slice` where `T: Copy`


zslayton:
When using a `Vec<'_, u8>`, I was surprised to see that `my_vec.extend_from_slice("SUCCESS".as_bytes());` produced assembly that looks like this on `x86_64`:

![image](https://github.com/fitzgen/bumpalo/assets/611616/5f2e657f-7f9e-4454-8ab2-26637987f21a)

_(mac OS 14.3 Sonoma, Intel processor, `rustc` v1.76.0)_

Notice that it's performing a reserve, error check, and push for each byte in the input slice. I had expected LLVM to reduce this to something like a single reserve, error check, and `memcpy` for the entire slice.

I'd like to contribute a change analogous to the one recently merged in #229 but for `&[u8]` instead of `&str`. Before I get started, I wanted to check whether 1) you'd be interested in merging such an optimization and 2) what the API should look like (since we don't have specialization to customize `extend_from_slice` for `Vec<'_, u8>`).

zslayton:
> What did you use to produce the visualization of assembly branch structure above?

I used [Cutter](https://cutter.re/), an open source reverse engineering tool. It's very helpful!

> Please note, we actually saw a very slight performance degredation from https://github.com/fitzgen/bumpalo/pull/229 in OXC https://github.com/oxc-project/oxc/pull/2417. I believe this is due to OXC only dealing with really short strings, where copying byte-by-byte (presumably inlined) is actually faster than a call to copy_nonoverlapping. I don't know if this should be addressed or not, as it's probably an unusual case (std does not address it). I should have made a benchmark for short strings as well as long.

That's good food for thought, thank you for raising it--I'll defer to @fitzgen as to whether it needs to be addressed. My use case is a mix of short strings and byte arrays that are anywhere from empty to several kilobytes in typical scenarios. The change in #236 was a very large performance increase in all of my existing benchmarks, but I didn't closely examine the performance of small slices.

zslayton:
@overlookmotel I ended up extending the benchmark to test a variety of input sizes. You can see the results [here](https://github.com/fitzgen/bumpalo/pull/236#issuecomment-1952764042).

zslayton:
This was fixed by #236.

PR #236, opened 2024-02-18
Provides implementation of `Vec::extend_from_slice` optimized for `T: Copy`

zslayton:
Addresses #235.

This PR is meant to be a jumping-off point for discussion. It adds a new inherent impl method to `Vec`:

```rust
impl<'bump, T: 'bump + Copy> Vec<'bump, T> {
    /// ...
    pub fn extend_from_slice_copy(&mut self, other: &[T]) {
    // ...
    }
}
```

It moves the logic added in #229 from `String` to `Vec` and then re-implements `push_str` by calling the new `extend_from_slice_copy`.

I've duplicated the `push_str` benchmarks but for `Vec<'_, u8>` and the performance improves as expected:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,2 @@
 target
 Cargo.lock
-
-# JetBrains IDE files (e.g. RustRover)
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 target
 Cargo.lock
+
+# JetBrains IDE files (e.g. RustRover)
+.idea
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -131,34 +131,46 @@ fn extend_from_slice_copy_u8(bump: &bumpalo::Bump, slice: &[u8]) {
 const ALLOCATIONS: usize = 10_000;
 
 fn bench_extend_from_slice_copy(c: &mut Criterion) {
-    let len: usize = 16 * 1024; // 16 KiB
+    let lengths = &[
+        4usize,
+        8,
+        16,
+        64,
+        128,
+        1024,
+        4 * 1024,
+        16 * 1024,
+    ];
 
-    let mut group = c.benchmark_group("extend");
-    group.throughput(Throughput::Elements(len as u64));
-    group.bench_function("extend", |b| {
-        let mut bump = bumpalo::Bump::with_capacity(len);
+    for len in lengths.iter().copied() {
         let str = "x".repeat(len);
-        b.iter(|| {
-            bump.reset();
-            extend_u8(&bump, str.as_bytes());
+        let mut group = c.benchmark_group(format!("extend {len} bytes"));
+        group.throughput(Throughput::Elements(len as u64));
+        group.bench_function("extend", |b| {
+            let mut bump = bumpalo::Bump::with_capacity(len);
+            b.iter(|| {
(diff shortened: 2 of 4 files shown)


zslayton:
@overlookmotel [expressed a concern](https://github.com/fitzgen/bumpalo/issues/235#issuecomment-1952335030) that unconditionally using `copy_nonoverlapping` might result in small performance regressions in the case of a small slices/strings.

I extended the benchmarks in this PR to test a variety of sizes between 4 bytes and 16KB. The results--in this microbenchmark, running on my machine--indicated that `copy_nonoverlapping` is a win for slices of any size.

image
image

overlookmotel:
That's interesting! I wonder why we saw a slow-down in OXC (though an *extremely* small one) with the other change.

The other explanations I can think of are:

1. `criterion::black_box` is not doing its job properly in your benchmarks, which would allow the compiler to convert the `copy_nonoverlapping` calls to inlined copies.
2. It performs better with slice lengths which are multiples of 8 / 16 etc (seems plausible). Maybe try a "weird" length like 5 or 7 and see how it does with that?

Sorry, not trying to pick holes in your work, just trying to understand.

.gitignore
```diff
@@ -1,2 +1,5 @@
 target
 Cargo.lock
+
+# JetBrains IDE files (e.g. RustRover)
+.idea
```
fitzgen, line 5: This seems like something that is more appropriate for your home directory's `.gitignore` than every project you contribute to.

fitzgen (commented):
Thanks! Looks good to me modulo nitpick below.

Regarding short strings and possible perf regressions: nothing I've read seems serious enough to warrant reverting that optimization. I'm happy to receive follow up PRs that further improve things here, if there is anything further to improve.

.gitignore
zslayton: 🤦 TIL that you can put a `.gitignore` in your home directory. I'll remove this, thanks.

I also noticed that `Vec`'s impl of `std::io::Write` can be updated to use this method; I'll push another commit with that change tomorrow.

overlookmotel:
> Regarding short strings and possible perf regressions: nothing I've read seems serious enough to warrant reverting that optimization. I'm happy to receive follow up PRs that further improve things here, if there is anything further to improve.

Yes absolutely, the performance "regression" we saw was absolutely tiny, and for general purpose use 80x faster speed for long strings outweighs 0.1% slower for very short strings. I just wanted to put it on the record that there was *some* downside to the PR I made, rather than leave it unsaid, in case it affects anyone else.

fitzgen (approved):
Thanks!

overlookmotel:
@zslayton Thanks for tolerating my nitty comments, and for benching shorter byte ranges. I remain curious about why we saw the slight slow-down we did on similar PR for strings, but probably I should stop worrying!

zslayton:
> @zslayton Thanks for tolerating my nitty comments, and for benching shorter byte ranges. 

Sure thing! I appreciate having another set of eyes on this, I'd rather spend some time double-checking things than accidentally cause a regression for [all of these folks](https://crates.io/crates/bumpalo/reverse_dependencies) 🙃😬.

> I remain curious about why we saw the slight slow-down we did on similar PR for strings, but probably I should stop worrying!

I'm curious too!

PR #239, opened 2024-02-21
Modifies RawVec `reserve` fn structure to improve inlining

zslayton:
This PR reorganizes `RawVec::reserve_internal` into a family of methods that allow the compiler to make better inlining decisions.

`fallible_reserve_internal` is a wrapper around `reserve_internal` that performs the "do we already have enough space?" check:
```rust
if self.cap().wrapping_sub(used_cap) >= needed_extra_cap {
    return Ok(());
}
```
This portion of the code (which produces a very small amount of assembly) is always inlined. Only if this check fails (which is relatively rare) does it then call `reserve_internal_or_error`, which is not inlined.

The `infallible_reserve_internal`/`reserve_internal_or_panic` method pair is similar to the above, but also prevents the code needed to detect an error and panic accordingly from being inlined.

Collectively, this allows the compiler to inline `Vec::push` and `Vec::extend*` in more places by transitively shrinking the amount of assembly each produces.

You can see the impact in the benchmarks introduced in #235:

image

### DIFF
--- a/src/collections/raw_vec.rs
+++ b/src/collections/raw_vec.rs
@@ -319,7 +319,7 @@ impl<'a, T> RawVec<'a, T> {
         used_cap: usize,
         needed_extra_cap: usize,
     ) -> Result<(), CollectionAllocErr> {
-        self.reserve_internal(used_cap, needed_extra_cap, Fallible, Exact)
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Looks great, thanks!

PR #240, opened 2024-03-01
Adds `Vec::extend_from_slices_copy` that accepts multiple slices

zslayton:
This PR builds on #236 by introducing an `extend_from_slices_copy` method.

The singular `extend_from_slice_copy` method calls `self.reserve()` before copying the input slice. In many circumstances, users will have several inputs to append at once; for example:

```rust
let mut vec = Vec::new_in(&bump);
vec.extend_from_slice_copy("foo".as_bytes()); // reserve() + memcpy
vec.extend_from_slice_copy("bar".as_bytes()); // reserve() + memcpy
vec.extend_from_slice_copy("baz".as_bytes()); // reserve() + memcpy
assert_eq!("foobarbaz".as_bytes(), vec.as_slice());
```

The compiler is often (always?) unable to optimize away the redundant calls to `reserve()`, which leads to assembly like the following:

image

If there is not sufficient space already allocated, it's possible for each intermediate call to `reserve` to cause a reallocation that could have been avoided if the total space required were known in advance.

The new `extend_from_slices_copy` method takes a slice-of-slices, allowing it to precompute the space required and perform a single call to `reserve`.

```rust
let mut vec = Vec::new_in(&bump);
vec.extend_from_slices_copy(&[ // reserve()
  "foo".as_bytes(), // memcpy
  "bar".as_bytes(), // memcpy
  "baz".as_bytes(), // memcpy
])
assert_eq!("foobarbaz".as_bytes(), vec.as_slice());
```

These benchmarks show the relative performance of the two methods when there are `N` slices to append and there is not sufficient memory allocated in advance:

image

Note that there is a minor performance penalty for using `extend_from_slices_copy` when there's only one input slice. 

In the case where sufficient space has already been allocated in the `Bump` and `Vec` ahead of time, `extend_from_slices_copy` is faster by a narrower margin.

<details>
<summary>preallocated=true benchmark results</summary>

image

</details>

Anecdotally, `extend_from_slices_copy` produces simpler assembly and the compiler also seems to be able to unroll the loops reliably when slice literals are being passed in (vs not-known-until-runtime and/or `black_box` slices, which is shown in the benchmarks above).

### DIFF
--- a/src/collections/vec.rs
+++ b/src/collections/vec.rs
@@ -1894,11 +1894,11 @@ impl<'bump, T: 'bump + Copy> Vec<'bump, T> {
     pub fn extend_from_slices_copy(&mut self, slices: &[&[T]]) {
         // Reserve the total amount of capacity we'll need to safely append the aggregated contents
         // of each slice in `slices`.
-        let capacity_to_reserve: usize = slices.iter().map(|buf| buf.len()).sum();
+        let capacity_to_reserve: usize = slices.iter().map(|slice| slice.len()).sum();
         self.reserve(capacity_to_reserve);
 
         // SAFETY:
-        // * `dst` is valid for writes of `other.len()` bytes as `self.reserve(other.len())`
+        // * `dst` is valid for writes of `other.len()` items as `self.reserve(other.len())`
         //   above guarantees that.
         // * Source and destination ranges cannot overlap as we just reserved the destination
         //   range from the bump.
--- a/src/collections/vec.rs
+++ b/src/collections/vec.rs
@@ -1782,13 +1782,15 @@ impl<'bump, T: 'bump + Copy> Vec<'bump, T> {
     ///
     /// SAFETY:
     ///   * The caller is responsible for:
-    ///       * calling `reserve` beforehand to guarantee that there is enough capacity to
-    ///         store `other.len()` more items.
-    ///       * updating the length of the `Vec` afterward to reflect the number of items added.
+    ///       * calling [`reserve`](Self::reserve) beforehand to guarantee that there is enough
+    ///         capacity to store `other.len()` more items.
+    ///       * updating the length of the `Vec` afterward to reflect the number of items
+    ///         added (usually via [`set_len`](Self::set_len)).
     ///   * The caller must also guarantee that `self` and `other` do not overlap.
     unsafe fn extend_from_slice_copy_unchecked(&mut self, other: &[T]) {
         let old_len = self.len();
         debug_assert!(old_len + other.len() <= self.capacity());
+        debug_assert!(self.capacity() - self.len() >= other.len());
 
         // SAFETY:
         // * `src` is valid for reads of `other.len()` values by virtue of being a `&[T]`.
@@ -1892,8 +1894,8 @@ impl<'bump, T: 'bump + Copy> Vec<'bump, T> {
     pub fn extend_from_slices_copy(&mut self, slices: &[&[T]]) {
         // Reserve the total amount of capacity we'll need to safely append the aggregated contents
         // of each slice in `slices`.
-        let bytes_to_reserve: usize = slices.iter().map(|buf| buf.len()).sum();
-        self.reserve(bytes_to_reserve);
+        let capacity_to_reserve: usize = slices.iter().map(|buf| buf.len()).sum();
+        self.reserve(capacity_to_reserve);
 
         // SAFETY:
         // * `dst` is valid for writes of `other.len()` bytes as `self.reserve(other.len())`
--- a/src/collections/vec.rs
+++ b/src/collections/vec.rs
@@ -1778,11 +1778,42 @@ impl<'bump, T: 'bump + Clone> Vec<'bump, T> {
 }
 
 impl<'bump, T: 'bump + Copy> Vec<'bump, T> {
+    /// Helper method to copy all of the items in `other` and append them to the end of `self`.
+    ///
+    /// SAFETY:
+    ///   * The caller is responsible for:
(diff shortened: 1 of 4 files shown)


src/collections/vec.rs
```diff
@@ -1778,12 +1778,43 @@ impl<'bump, T: 'bump + Clone> Vec<'bump, T> {
...
+    ///   * The caller must also guarantee that `self` and `other` do not overlap.
+    unsafe fn extend_from_slice_copy_unchecked(&mut self, other: &[T]) {
+        let old_len = self.len();
+        debug_assert!(old_len + other.len() <= self.capacity());
+
+        // SAFETY:
+        // * `src` is valid for reads of `other.len()` values by virtue of being a `&[T]`.
+        // * `dst` is valid for writes of `other.len()` bytes because the caller of this
+        //   method is required to `reserve` capacity to store at least `other.len()` items
+        //   beforehand.
```
fitzgen, line 1797: Can we debug assert this property? Something like

```
debug_assert!(self.capacity() - self.len() >= other.len());
```

src/collections/vec.rs
```diff
@@ -1778,12 +1778,43 @@ impl<'bump, T: 'bump + Clone> Vec<'bump, T> {
...
+        // * Because `src` is a `&[T]` and dst is a `&[T]` within the `Vec<T>`,
+        //   `copy_nonoverlapping`'s alignment requirements are met.
+        // * Caller is required to guarantee that the source and destination ranges cannot overlap
+        unsafe {
+            let src = other.as_ptr();
+            let dst = self.as_mut_ptr().add(old_len);
+            ptr::copy_nonoverlapping(src, dst, other.len());
+        }
+    }
```
fitzgen, line 1806: This function doesn't update `self.len`, but I think it should, no?

src/collections/vec.rs
```diff
@@ -1806,35 +1837,79 @@ impl<'bump, T: 'bump + Copy> Vec<'bump, T> {
...
+    /// vec.extend_from_slices_copy(&["ello,".as_bytes(), &[], " world!".as_bytes()]);
+    /// assert_eq!(vec, "Hello, world!".as_bytes());
+    /// ```
+    ///
+    /// [`extend_from_slice_copy`]: #method.extend_from_slice_copy
+    pub fn extend_from_slices_copy(&mut self, slices: &[&[T]]) {
+        // Reserve the total amount of capacity we'll need to safely append the aggregated contents
+        // of each slice in `slices`.
+        let bytes_to_reserve: usize = slices.iter().map(|buf| buf.len()).sum();
+        self.reserve(bytes_to_reserve);
```
fitzgen, line 1896: This is not actually bytes, its elements, right? Should update the variable names accordingly.

src/collections/vec.rs
```diff
@@ -1806,35 +1837,79 @@ impl<'bump, T: 'bump + Copy> Vec<'bump, T> {
...
-            let src = other.as_ptr();
-            let dst = self.as_mut_ptr().add(old_len);
-            ptr::copy_nonoverlapping(src, dst, other.len());
+            // Copy the contents of each slice onto the end of `self`
+            slices.iter().for_each(|slice| {
+                self.extend_from_slice_copy_unchecked(slice);
+                // Update the Vec's length to reflect the number of items we just added. This needs
+                // to happen for each slice so the next call to `extend_from_slice_copy_unchecked`
+                // will write to the correct (updated) offset.
+                self.set_len(self.len() + slice.len());
```
fitzgen, line 1910: I guess all callers of `extend_from_slice_copy_unchecked` are required to update lengths themselves. Can we document that in the method's docs as another thing that is the caller's responsibility?

fitzgen (commented):
Thanks!

All this unsafe code makes it so that I'd like some quickchecks to better exercise edge cases and give confidence that all t's are crossed and i's are dotted. Could you write the following quickchecks:

* a quickcheck for `extend_from_slice_copy` that takes two slices, creates a vec from the first one, extends with the second, and asserts that the resulting vec length and contents are as expected
* a similar quickcheck for `extend_from_slices_copy` that takes a list of slices, rather than just two, and creates a vec of the first one and then calls `extend_from_slices_copy` with the rest of them and checks that the resulting vec length and contents are as expected
* a differential quickcheck: take a list of slices, and do two things, (1) take the first slice and turn it into a vec, then call `extend_from_slice_copy` n times with each of the rest of the slices, and (2) take the first slice again and turn it into another vec, then call `extend_from_slices_copy` once with all the rest of the slices. Finally, it should check that both vecs have the same length and contents.

src/collections/vec.rs
zslayton: This is the caller's responsibility, which allows this method to be used for both the single-slice and multi-slice cases. There's a note about this in the doc comment above:

```rust
/// * The caller is responsible for:
///   * calling `reserve` beforehand to guarantee that there is enough capacity to
///     store `other.len()` more items.
///   * updating the length of the `Vec` afterward to reflect the number of items added.
```

I'll add an explicit note about calling `set_len` to that last bullet. Please let me know if you'd like me to do something else to make it clearer.

src/collections/vec.rs
zslayton: I added this `debug_assert!` and then later realized that it was equivalent to the one already on line 1790 above the `SAFETY` comment. I removed it again in commit https://github.com/fitzgen/bumpalo/pull/240/commits/60b1407608fdc1c98bf64477ee5bc10facd692fb.

fitzgen (commented):
Looks great, thanks!

fitzgen (approved):

PR #248, opened 2024-05-19
Fixed stacked borrows violation in `dealloc` (fixes #247)

### LINKED ISSUE #247: Miri error with `allocator_api` and `Box`-pointer roundtrips
Hi, thank you for the great work!

Using the `allocator_api` feature of this crate and nightly, I noticed that Miri does not like this:

```rust
#![feature(allocator_api)]

fn main() {
    let bump = bumpalo::Bump::new();

    let a = Box::into_raw(Box::new_in(1u8, &bump));
    drop(unsafe { Box::from_raw_in(a, &bump) });

    let _b = Box::new_in(2u16, &bump);
}
```

Miri error:

```
error: Undefined Behavior: trying to retag from <2737> for SharedReadWrite permission at alloc1066[0x1be], but that tag does not exist in the borrow stack for this location
   --> C:\...\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\alloc\src\boxed.rs:385:13
    |
385 |             boxed.as_mut_ptr().write(x);
    |             ^^^^^
    |             |
    |             trying to retag from <2737> for SharedReadWrite permission at alloc1066[0x1be], but that tag does not exist in the borrow stack for this location
    |             this error occurs as part of two-phase retag at alloc1066[0x1be..0x1c0]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <2737> was created by a SharedReadWrite retag at offsets [0x1bf..0x1c0]
   --> src/main.rs:6:13
    |
6   |     let a = Box::into_raw(Box::new_in(1u8, &bump));
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: BACKTRACE (of the first span):
    = note: inside `std::boxed::Box::<u16, &bumpalo::Bump>::new_in` at C:\...\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\alloc\src\boxed.rs:385:13: 385:18
note: inside `main`
   --> src/main.rs:9:14
    |
9   |     let _b = Box::new_in(2u16, &bump);
    |              ^^^^^^^^^^^^^^^^^^^^^^^^
```

The exact types of the allocations involved does not matter much, only that `_b` allocates something with an alignment larger than `a`.

I'm pretty new to Miri, so I'm not quite sure what the issue is exactly. It seems like the combination of the retagging in `Box::into_raw`, the deallocation in of the recreated box, and the layout differences are the essential parts. This is not triggered if the `Box` is dropped without the pointer roundtrip, nor with `bumpalo::boxed::Box`, nor when using `bump.reset()` instead of `drop`. It does show up when using `allocator-api2` and its `Box`.

bluurryy:
In some strange edge case miri complains about a stacked borrows violation (see #247). I don't understand why the violation gets triggered in this particular case.

But when using the pointer from the footer instead of using the parameter provided one, miri won't complain anymore.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1705,6 +1705,7 @@ impl Bump {
         // If the pointer is the last allocation we made, we can reuse the bytes,
         // otherwise they are simply leaked -- at least until somebody calls reset().
         if self.is_last_allocation(ptr) {
+            let ptr = self.current_chunk_footer.get().as_ref().ptr.get();
             let ptr = NonNull::new_unchecked(ptr.as_ptr().add(layout.size()));
             self.current_chunk_footer.get().as_ref().ptr.set(ptr);
         }
--- a/tests/all/tests.rs
+++ b/tests/all/tests.rs
@@ -209,3 +209,14 @@ fn test_chunk_capacity() {
     b.alloc(true);
     assert!(b.chunk_capacity() < orig_capacity);
 }
+
+#[test]
+#[cfg(feature = "allocator_api")]
+fn miri_stacked_borrows_issue_247() {
+    let bump = bumpalo::Bump::new();
+
+    let a = Box::into_raw(Box::new_in(1u8, &bump));
+    drop(unsafe { Box::from_raw_in(a, &bump) });
+
+    let _b = Box::new_in(2u16, &bump);
+}


fitzgen (commented):
Interesting. Probably a provenance thing.

Can you add a regression test? Thanks!

fitzgen (approved):
Thanks!

PR #249, opened 2024-05-19
Remove bogus executable filesystem permissions from Rust sources

musicinmybrain:
It does not make sense for non-script files to have the executable bit set in their filesystem permissions.

In particular, executable Rust sources can cause minor problems with Linux distribution packaging because `#![…]` directives look like shebang lines, and this can confuse automation.


fitzgen (approved):
Thanks! Yeah emacs will helpfully do this automatically sometimes if a file starts with `#!`

PR #252, opened 2024-05-19
Reduce rust-version to 1.71.1

dahlbaek:
As far as I can tell, the crate compiles fine against rustc 1.71.1, which is the rustc version of Debian trixie: https://packages.debian.org/testing/rustc. Thus, for the benefit of distro packagers and others that might be using old versions of rustc, it would be neat if rust-version could be reduced to 1.71.1.


fitzgen (approved):
Huh, at some point I think we used something from 1.73, but I guess not anymore. Seems fine to lower it to 1.71.1 now though, since all tests are already passing there!

PR #253, opened 2024-06-07
Add integration with zerocopy crate

sivadeilra:
The zerocopy crate provides traits and functions for _safely_ transmuting types to/from bytes. It is an important building block for high-performance serialization in many designs. It is also a very well-maintained crate with high standards for quality and its maintainers are actively engaged with the Rust Project on advancing the goals of the Safe Transmute working group.

The `zerocopy` crate defines the `FromZeroes` trait, which specifies that a type can be safely constructed from a buffer containing an all-zeroes bit pattern.

This PR adds two new methods to `Bump`: `new_zeroed` and `new_slice_zeroed`. `new_zeroed` allocates space for `T` (where `T: FromZeroes`) and returns `&mut T`. This avoids the "placement new" problem in Rust, which causes problems when attempting to allocate large types in the heap, such as `[u8; 0x10000]`. For most containers, such as `Box`, the type is briefly constructed in a temporary on the stack, then a heap allocation is done, then the value is moved into the heap allocation. If `T` is large enough, it can cause stack overflow.

The `Bump::new_zeroed` function avoids this problem by simply allocating space for `T` directly in the heap, then filling it with zeroes, then casting the allocation as `&mut T` and returning it. The `FromZeroes` constraint ensures that this is sound.

The `new_slice_zeroed` function similarly allows allocating slices directly in a Bump.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,11 @@ harness = false
 # implement its `Allocator` trait.
 allocator-api2 = { version = "0.2.8", default-features = false, optional = true }
 
+# This dependency enables integration with the zerocopy trait, which allows
+# allocating values that implement FromZeroes directly in a Bump.
+zerocopy = { version = "0.7.34", default-features = false, optional = true }
+zerocopy-derive = { version = "0.7.34", default-features = false, optional = true }
+
 # This dependency is here to allow integration with Serde, if the `serde` feature is enabled
 serde = { version = "1.0.171", optional = true }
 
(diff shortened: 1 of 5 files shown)


fitzgen (commented):
Happy to take a PR that adds a zeroing allocation method, but I don't want to take a dependency on any new crates.

sivadeilra:
> Happy to take a PR that adds a zeroing allocation method, but I don't want to take a dependency on any new crates.

It's an optional dependency, though, and the `zerocopy` crate is part of the Safe Transmute initiative.

PR #260, opened 2024-12-26
add `try_` functions for slices and `str`

c-cube:
It's good to be able to deal with allocation errors explicitly.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1486,7 +1486,7 @@ impl Bump {
         })
     }
 
-    /// Allocates a new slice of size `len` slice into this `Bump` and return an
+    /// Allocates a new slice of size `iter.len()` slice into this `Bump` and return an
     /// exclusive reference to the copy. Does not panic on failure.
     ///
     /// The elements are initialized using the supplied iterator.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1180,6 +1180,34 @@ impl Bump {
         }
     }
 
+    /// Like `alloc_slice_copy`, but does not panic in case of allocation failure.
+    ///
+    /// ## Example
+    ///
+    /// ```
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1180,6 +1180,34 @@ impl Bump {
...
+    /// assert_eq!(x, Err(bumpalo::AllocErr)); // too big
+    /// ```
+    #[inline(always)]
+    pub fn try_alloc_slice_copy<T>(&self, src: &[T]) -> Result<&mut [T], AllocErr>
+    where
+        T: Copy,
+    {
+        let layout = Layout::for_value(src);
+        let dst = self.try_alloc_layout(layout)?.cast::<T>();
+        let result = unsafe { slice::from_raw_parts_mut(dst.as_ptr(), src.len()) };
```
fitzgen, line 1205: I don't think this is safe because it is creating a reference of potentially uninitialized data. I think this needs to use [`std::ptr::copy_nonoverlapping`](https://doc.rust-lang.org/nightly/std/ptr/fn.copy_nonoverlapping.html) instead.

src/lib.rs
```diff
@@ -1354,6 +1486,31 @@ impl Bump {
         })
     }
 
+    /// Allocates a new slice of size `len` slice into this `Bump` and return an
```
fitzgen, line 1489:
```suggestion
    /// Allocates a new slice of size `iter.len()` slice into this `Bump` and return an
```

fitzgen (commented):
Thanks! And sorry about the delay. This looks good, but just a few little comments below.

src/lib.rs
c-cube: Thank you, good point. `core::ptr::copy_nonoverlapping` for the no_std case.

fitzgen (approved):
Thanks!

PR #264, opened 2025-02-12
use primitive type u64 in README

ArtemIsmagilov:
Hi!
module [std::u64](https://doc.rust-lang.org/std/u64/index.html) is deprecated.
method [max_value](https://doc.rust-lang.org/std/primitive.u64.html#method.max_value) also deprecated


fitzgen (approved):
Thanks!

ArtemIsmagilov:
And you too!

PR #266, opened 2025-02-24
Add serde serialization for String

aabizri:
Previously (#210 and d746a56a51b73a53eb666f1a29961466ab84478a), `serde` support was added to `Vec` and `Box` but not `String`. This fills that gap. 

Additionally reorganize tests to split the ones that depend on `collections` and `boxed` features, as before, running `cargo test --features=boxed,serde` didn't run the tests for `Box` as they were under a `collections` feature guard.

Related to #63


fitzgen (approved):
Thanks!

PR #275, opened 2025-06-05
fix: subtract footer size from allocated_bytes on reset

### LINKED ISSUE #274: allocated_bytes after reset erroneously includes footer size
New chunks don't include the footer size when created, but the chunk after a `reset()` does

```rust
#[test]
fn allocated_bytes() {
    let mut b = Bump::with_capacity(1);

    assert_eq!(b.allocated_bytes(), DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER);
    assert_eq!(
        b.allocated_bytes_including_metadata(),
        DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER + FOOTER_SIZE
    );

    b.reset();

    assert_eq!(b.allocated_bytes(), DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER);
    assert_eq!(
        b.allocated_bytes_including_metadata(),
        DEFAULT_CHUNK_SIZE_WITHOUT_FOOTER + FOOTER_SIZE
    );
}
```

This fails on the first assert AFTER the `reset()`

```bash
assertion `left == right` failed
  left: 496
 right: 448
```

joe-p:
When a new chunk is created, allocated_bytes does not include the footer size. In reset, however, the footer size was implicitly included. This subtracts the footer size from the chunks new allocated_bytes

Fixes #274

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -990,7 +990,7 @@ impl<const MIN_ALIGN: usize> Bump<MIN_ALIGN> {
             cur_chunk.as_ref().ptr.set(cur_chunk.cast());
 
             // Reset the allocated size of the chunk.
-            cur_chunk.as_mut().allocated_bytes = cur_chunk.as_ref().layout.size();
+            cur_chunk.as_mut().allocated_bytes = cur_chunk.as_ref().layout.size() - FOOTER_SIZE;
 
             debug_assert!(
                 self.current_chunk_footer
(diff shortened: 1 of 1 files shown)


joe-p:
This is just for a side project so not urgent. For some context this came up when checking if the allocator was reset before it's used in a function that's called in a loop. I was originally checking `used_capacity == 0`, but I can just make sure it has not exceeded the footer for now. Of course this is fragile to footer size changing, but good enough for now until you're ready to release. 

```rust
let used_capacity = arena.allocated_bytes() - arena.chunk_capacity();

if used_capacity > 6 * mem::size_of::<usize>() {
    panic!("Bump allocator must be reset");
}
```

fitzgen:
Published `3.18.0`

Issue #277, opened 2025-06-05
error building regalloc2-0.12.2


morenol:
```
--> /Users/morenol/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/regalloc2-0.12.2/src/lib.rs:1698:24
     |
1698 |         self.0.deref().allocate(layout)
     |                        ^^^^^^^^
     |
    ::: /Users/morenol/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bumpalo-3.18.0/src/lib.rs:292:1
     |
292  | pub struct Bump<const MIN_ALIGN: usize = 1> {
     | ------------------------------------------- doesn't satisfy `bumpalo::Bump: allocator_api2::alloc::Allocator`
     |
     = note: the following trait bounds were not satisfied:
             `bumpalo::Bump: allocator_api2::alloc::Allocator`
             which is required by `&bumpalo::Bump: allocator_api2::alloc::Allocator`
     = help: items from traits can only be used if the trait is in scope
help: trait `Allocator` which provides `allocate` is implemented but not in scope; perhaps you want to import it
     |
17   + use allocator_api2::alloc::Allocator;
```

Seems to be related to this release?

jadamcrain:
@bjorn3 Ah!! Thanks for that.

fitzgen:
Ah, I was under the impression that it was doing the re-export trick. Will yank, revert the `allocator_api2` bump, and publish 3.18.1.

fitzgen:
Revert in https://github.com/fitzgen/bumpalo/pull/278

PR #288, opened 2025-07-16
Bump allocator-api2 to 0.3.0

gagbo:
I do not know how breaking that change is, so I’ll just highlight that I did _not_ touch
the version in `Cargo.toml`


gagbo:
Also, I know it’s an extremely small PR, feel free to close it/remove it if necessary; I just want to provide a quick one click solution if the change seems ok to you

fitzgen:
Thanks for making this PR.

Because `allocator_api2` is a public dependency, this can't merge until the next breaking change. See https://github.com/fitzgen/bumpalo/issues/277 for some details and discussion.

clarfonthey:
Presumably this is redundant to #299.

PR #290, opened 2025-09-25
Add inline attribute to String::from_str_in

nickbabcock:
Saw that `String::from_str_in` pop up in my own application benchmark profiling. It seems like a good candidate for inlining as `String::with_capacity_in` is inlined and `String::from_str_in` isn't doing much work other than that.

Running the bumpalo benchmarks, the `from_str_in` benchmark did not see any changes, though my own benchmarks show a consistent 5-10% improvement in latency.

### DIFF
--- a/src/collections/string.rs
+++ b/src/collections/string.rs
@@ -679,6 +679,7 @@ impl<'bump> String<'bump> {
     /// let s = String::from_str_in("hello", &b);
     /// assert_eq!(s, "hello");
     /// ```
+    #[inline]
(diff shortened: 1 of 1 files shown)


fitzgen:
Thanks! Sorry for the delayed response. Need to figure out what to do about deps breaking MSRV.

fitzgen:
CI fixed in https://github.com/fitzgen/bumpalo/pull/292

@nickbabcock care to rebase so that CI retriggers and has those fixes? Thanks

fitzgen (approved):
Thanks!

PR #299, opened 2025-12-11
Update allocator-api2

zakarumych:
New version fixes some inconsistencies

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ harness = false
 # This dependency provides a version of the unstable nightly Rust `Allocator`
 # trait on stable Rust. Enabling this feature means that `bumpalo` will
 # implement its `Allocator` trait.
-allocator-api2 = { version = "0.2.8", default-features = false, optional = true }
+allocator-api2 = { version = "0.4.0", default-features = false, optional = true }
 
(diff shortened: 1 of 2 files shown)


fitzgen (commented):
This is a breaking change because `allocator_api2` is in the public API and this isn't a patch bump, correct?

clarfonthey:
Is there anything in particular you'd like before releasing a new breaking release? Since this repo doesn't seem to be particular active at the moment, I'm mostly wondering because this is technically blocking updating `allocator-api2` on repos that depend on `bumpalo`, like `hashbrown`.

The only thing that would particularly come to mind is *maybe* updating to edition 2024, but that probably isn't particularly desired since I assume that keeping an older MSRV is preferable.

fitzgen:
> Is there anything in particular you'd like before releasing a new breaking release?

There are a number of things I'd like to roll up together in the next breaking release to avoid needing to do multiple breaking releases: https://github.com/fitzgen/bumpalo/issues?q=state%3Aopen%20label%3Abreaking-change

PaulDance:
Would it be possible to make this move forward now?

xTachyon:
Hello. Can we please get a release with `allocator-api2` updated?

Issue #170, opened 2022-06-27
Try add covariance for containers


ZetaNumbers:
Enables a small(?) but valid set of cases. May come with a small penalty, until we have `Unique` built-in. Tho it seems like even in std `Unique<T>` does not enable appropriate optimizations currently, based on my experiments with `std::vec::Vec`. On the other hand `Box` enables "ownership optimizations", maybe due to `#[lang = "..."]` attribute.

Make sure to benchmark difference.

fitzgen:
I doubt that lifetime variance has any effect on optimizations. AFAIK that's all erased by the time code goes to LLVM, and I'd expect that any `rustc` optimizations for `Box` are looking at the lang attributes.

ZetaNumbers:
> I doubt that lifetime variance has any effect on optimizations.

Of course, it's just that there's currently no nonaliased covariant pointer for rust, so the only option is maybe aliased covariant `NonNull<T>`, vs. nonaliased but invariant `&mut ManuallyDrop<T>`

douglas-raillard-arm:
I just bumped into this issue while replacing a `std::boxed::Box` inside a recursive enum like this one:
```
struct Value<'a> {
    Str(&'a str),
    Wrapped(Box<Value<'a>>)
} 
```

The consuming code rightfully expects `Value<'a>` to be covariant in `'a`, but that breaks with bumpalo currently.

PR #304, opened 2026-02-13
Make box covariant

ElchananHaas:
This CR makes the box covariant in 'a to allow for more use cases. This addresses #170.

Making it covariant should be OK for the same reasons as the standard library box is safe. Quoting the RustNomicon (https://doc.rust-lang.org/nomicon/subtyping.html)

> As it turns out, the argument for why it's ok for Box (and Vec, HashMap, etc.) to be covariant is pretty similar to the argument for why it's ok for lifetimes to be covariant: as soon as you try to stuff them in something like a mutable reference, they inherit invariance and you're prevented from doing anything bad.

The Bumpalo Vec is already covariant. 

I tested this code by running

    cargo test --features boxed 

I checked that all tests passed. I also added a unit test for covariance. I wasn't able to come up with a great test, but I verified that the test fails if I make the lifetime invariant. 

I additionally ran 

    cargo miri test --features boxed

miri passed all tests. It complains about some memory leaks in raw_vec, but that should be unrelated.

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -121,22 +121,22 @@
 
 use {
     crate::Bump,
-    {
-        core::{
-            any::Any,
-            borrow,
-            cmp::Ordering,
-            convert::TryFrom,
-            future::Future,
-            hash::{Hash, Hasher},
-            iter::FusedIterator,
-            mem::ManuallyDrop,
-            ops::{Deref, DerefMut},
-            pin::Pin,
-            task::{Context, Poll},
-        },
-        core_alloc::fmt,
+    core::{
+        any::Any,
+        borrow,
+        cmp::Ordering,
+        convert::TryFrom,
+        future::Future,
+        hash::{Hash, Hasher},
+        iter::FusedIterator,
+        marker::PhantomData,
+        mem::ManuallyDrop,
+        ops::{Deref, DerefMut},
+        pin::Pin,
+        ptr::NonNull,
+        task::{Context, Poll},
     },
+    core_alloc::fmt,
 };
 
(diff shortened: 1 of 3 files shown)


ElchananHaas:
The build is failing 

      
      [0](https://github.com/fitzgen/bumpalo/actions/runs/21973607396/job/63480335544?pr=304#step:5:21)
      error: failed to download replaced source registry `crates-io`
      
      Caused by:
        failed to parse manifest at `/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rand_core-0.10.0/Cargo.toml`
      
      Caused by:
        failed to parse the `edition` key
      
      Caused by:
        this version of Cargo is older than the `2024` edition, and only supports `2015`, `2018`, and `2021` editions.
      Error: Process completed with exit code 101.

I think this is unrelated.

src/boxed.rs
```diff
@@ -551,13 +555,15 @@ impl<'a, T: ?Sized> Deref for Box<'a, T> {
...
-        &*self.0
+        //Safety - The box points to a valid instance of T allocated with a Bumpalo arena.
+        unsafe { self.0.as_ref() }
     }
 }
 
 impl<'a, T: ?Sized> DerefMut for Box<'a, T> {
     fn deref_mut(&mut self) -> &mut T {
-        self.0
+        //Safety - The box points to a valid instance of T allocated with a Bumpalo arena.
```
fitzgen: Minor style nitpick:

```suggestion
        // Safety: The box points to a valid instance of T allocated with a Bumpalo arena.
```

PR #319, opened 2026-04-02
refactor: remove `bump` field from `ChunkIter`

overlookmotel:
Minor simplification.

It's unnecessary for `ChunkIter` to contain a `PhantomData<&'a mut Bump>` because it also contains a `ChunkRawIter<'a, MIN_ALIGN>` - and *that* contains `PhantomData<&'a mut Bump>`. This is sufficient to "hold" the lifetime for both iterators.


overlookmotel:
Copilot's critique sounds reasonable! I'm not actually sure he's *right*, but keeping the `bump` field is more explicit. Closing.

PR #322, opened 2026-04-27
Fix memory leak in `alloc_try_with` and `try_alloc_try_with`

Shnatsel:
When allocating a new chunk, these functions erroneously mark it as full. The pointers are set to the start of the chunk, but the bump allocator grows downwards, so the space in the newly allocated chunk appears to be exhausted.

This doc comment might have been the source of the error, since it's not entirely clear what "start" and "end" means for an allocator growing downwards in the address space:

https://github.com/fitzgen/bumpalo/blob/657caf85e7984b0aa8bc114c09d9cc2afeaf240d/src/lib.rs#L301-L304

The issue was identified by Claude Opus 4.7 as part of a memory safety audit. It turned up no security issues as of commit 657caf85e7984b0aa8bc114c09d9cc2afeaf240d.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1245,7 +1245,7 @@ impl<const MIN_ALIGN: usize> Bump<MIN_ALIGN> {
                         // Because this is the only allocation in this chunk,
                         // we can reset the chunk's bump finger to the start of
                         // the chunk.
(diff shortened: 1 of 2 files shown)


Shnatsel:
Ah, it seems I've reinvented #315

PR #329, opened 2026-07-07
Replace match with method call in round_up_to_unchecked

hamzader1:
Replace the `match` expression in `round_up_to_unchecked` with the equivalent method call.


fitzgen:
Why?

I don't know if you are trying to farm contributions or this is just an LLM acting on its own or what, but it doesn't feel like it was worth my time as a maintainer to look at this PR

PR #332, opened 2026-07-09
add resize_with

### LINKED ISSUE #331: implement ``resize_with``
it's kind of surprising these 2 methods aren't implemented, and it should be fairly straightforward to implement it. would copying the existing impls from std be okay?

drac-dot-gay:
i opted to just use a wrapper struct for fns, probably good enough

closes #331

### DIFF
--- a/src/collections/vec.rs
+++ b/src/collections/vec.rs
@@ -886,6 +886,48 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
         unsafe { slice::from_raw_parts_mut(ptr, len) }
     }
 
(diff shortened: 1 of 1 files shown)


fitzgen:
I am in the process of getting a new breaking release out the door, and I plan on removing the collections module in favor of `allocator_api2::{vec,collections}`. Its `Vec` type already has `resize_with`, fyi. In the meantime, `bumpalo::collections` is considered frozen.

drac-dot-gay:
ah, that's totally understandable. best of luck!

PR #334, opened 2026-07-22
Implement Allocator for Bump, not &Bump

### LINKED ISSUE #283: Implement `Allocator` for `Bump`, not `&Bump`
> `Bump` does not implement `Allocator` but `&Bump` does. So to get a `&dyn Allocator` you'd have to take a reference to the reference:
> ```rust
> &&bump_arena as &dyn Allocator
> ```
> That being said, I think the `impl Allocator for &Bump` could be changed to `impl Allocator for Bump` so you wouldn't have that double indirection. 

 _Originally posted by @bluurryy in [#280](https://github.com/fitzgen/bumpalo/issues/280#issuecomment-2998098983)_

maxdexh:
Hey, I'm helping with the current work on the `allocator_api` feature and your crate is the most commonly used example of an `Allocator` implementor outside `std`, so thank you for writing it!

Generally the allocator api is written to work with owning allocators as well as static allocators (like `System`).
This means that you can freely implement `Allocator` for `Bump`, instead of for `&Bump`.
Doing so is sound because allocator implementors are explicitly allowed to clear their allocations like you do.

This will also allow your allocator to be used inside `Box` and `Arc`, i.e. `Arc<Bump>` can be used as the allocator for long-lived allocations.

The existing implementation for `&Bump` follows from the blanket impl in `std` and `allocator-api2`.

Fixes #283

### DIFF
--- a/src/emplace.rs
+++ b/src/emplace.rs
@@ -536,7 +536,7 @@ where
         let new_layout = Layout::array::<T>(new_capacity).map_err(|_| AllocErr)?;
 
         let ptr = self.inner.guard.ptr;
-        let new_ptr = unsafe { Bump::grow(self.bump(), ptr, old_layout, new_layout)? };
+        let new_ptr = unsafe { Bump::grow_internal(self.bump(), ptr, old_layout, new_layout)? };
 
         self.inner.guard.ptr = new_ptr;
         self.inner.ptr = new_ptr.cast();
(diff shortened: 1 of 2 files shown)


fitzgen (approved):
Thanks! I vaguely remember not fixing this earlier because it might be a breaking change, but (a) I can't remember why it would be, and (b) I'm in a slow process of making a new breaking release now anyways so it doesn't matter anymore.

PR #338, opened 2026-09-13
impl core::error::Error for AllocErr

### LINKED ISSUE #295: Is there a reason that AllocErr doesn't implement Error
It would be nice to be able to use `?` on it more naturally.

Techcable:
This was stabilized in Rust 1.81, so works fine with our MSRV of 1.91.

Fixes #295


fitzgen (commented):
Thanks!

fitzgen (approved):

